PR #1, opened 2020-11-10
Set up basic interactivity and models

kwunkiuma:
Added foundations for a conversation with the bot. ?event now causes the bot to initiate a conversation and will only react to a subset of responses.

Added models for storing necessary data.


DiscordBot/Commands/EventCommands.cs
```diff
@@ -0,0 +1,40 @@
...
+                "``edit`` - edit event.\n"
+            );
+
+            var interactivity = context.Client.GetInteractivityModule();
+            var response = await interactivity.WaitForMessageAsync(
+                message =>
+                    message.Content.ToLower() == "create" ||
+                    message.Content.ToLower() == "rm" ||
+                    message.Content.ToLower() == "edit" ||
+                    message.Content.ToLower() == "stop",
```
FridaTveit, line 27: I would extract this to a small function so that you can create a variable for the lowercase value. I'd probably also store the strings as private constants 🙂

DiscordBot/Models/DiscordEvent.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace DiscordBot.Models
+{
+    class DiscordEvent
+    {
+        public string Name { get; }
+        public string Description { get; }
+        public int Key { get; }
```
FridaTveit, line 10: Is this used? It doesn't appear to be assigned to 🙂

DiscordBot/Models/Signup.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace DiscordBot.Models
+{
+    class Signup
+    {
+        public string Id;
+        public string Role;
```
FridaTveit, line 6: Any reason not to add getters only for these and to set them through a constructor? 🙂

DiscordBot/Program.cs
```diff
@@ -1,21 +1,29 @@
...
+        private static InteractivityModule interactivity;
+
         private static void Main(string[] args)
         {
-            MainAsync().GetAwaiter().GetResult();
+            MainAsync(args).ConfigureAwait(false).GetAwaiter().GetResult();
         }
 
-        private static async Task MainAsync()
+        private static async Task MainAsync(string[] args)
```
FridaTveit, line 19: Are these arguments used? 🙂

FridaTveit (commented):
Nice! 😄 How does this change the functionality of the bot? Does it work as an echo bot now? 🙂

DiscordBot/Program.cs
kwunkiuma: Not currently used but was in the example code for the older documentation that we've now switched to.

FridaTveit: If it's not used then I'd remove it. You can always add it later if you end up needing it 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: Should this be in a helper command?

FridaTveit: I think just a private method here is fine 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -36,5 +43,10 @@ public async Task Event(CommandContext context)
                 await context.RespondAsync(response.Message.Content);
             }
         }
+
+        private bool IsValidResponse(string response, IEnumerable<string> validResponses)
```
FridaTveit, line 47: There's not really any reason to pass in the responses here, you can just use the class variable 🙂

DiscordBot/Models/Signup.cs
```diff
@@ -4,5 +4,11 @@ class Signup
     {
         public string Id;
         public string Role;
```
FridaTveit, line 6: You need to add a getter here, otherwise they can be publically changed by default 🙂

DiscordBot/Models/DiscordEvent.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace DiscordBot.Models
+{
+    class DiscordEvent
```
FridaTveit, line 6: Is this class actually used anywhere? 🙂

FridaTveit (commented):

DiscordBot/Models/DiscordEvent.cs
kwunkiuma: This class is to establish the structure for which methods in Benji's data access service should return.

FridaTveit: Okay. For future reference I would in general expect that to be defined when it should be used, e.g. when defining the service class interface. Otherwise it's easy to forget about the class and for it not to be noticed that it's unused 🙂

DiscordBot/Program.cs
```diff
@@ -1,13 +1,19 @@
...
+using DSharpPlus.CommandsNext;
+using DSharpPlus.Interactivity;
 
 namespace DiscordBot
 {
     internal class Program
     {
-        private static void Main(string[] args)
+        private static CommandsNextModule commands;
+        private static InteractivityModule interactivity;
```
FridaTveit, line 12: Why are these stored as static variables? It's best to avoid non constant static variables wherever possible, as it makes the code less predictable. I can't see any reason why these need to be stored as variables? 🙂

FridaTveit (commented):
One small thing I hadn't noticed before, otherwise looks good 🙂

DiscordBot/Models/DiscordEvent.cs
kwunkiuma: I expect Benji to need the class quite soon, should I leave it or remove it?

DiscordBot/Program.cs
kwunkiuma: That was the approach in the example, it seems to work without setting them as static variables so I'll do that for now.

DiscordBot/Models/DiscordEvent.cs
FridaTveit: Fine to leave, just a note for next time 🙂

FridaTveit:
@benjimarshall there's a note on this PR saying you force pushed the main branch? Is that right or has github got confused? 🙂

.github/workflows/main_discordbotsoftwire.yml
```diff
@@ -0,0 +1,37 @@
+# Docs for the Azure Web Apps Deploy action: https://github.com/Azure/webapps-deploy
```
FridaTveit, line 1: Should this be included in this PR? 🙂

FridaTveit (commented):
One question, looks like some git stuff might have gone wrong. Let me know if you need any help. Otherwise looks good 🙂

.github/workflows/main_discordbotsoftwire.yml
kwunkiuma: I think that sneaked in here when I merged main into this PR, but now main has been reverted so I'll get rid of that.

FridaTveit: Main has been reverted? Why?? There should never be any force push to main.

benjimarshall:
> @benjimarshall there's a note on this PR saying you force pushed the main branch? Is that right or has github got confused? 🙂

Yeah Shiv (@SChandarana) asked me to. All of the previous commits he had made directly to `main` were not going to be used any more, and he wanted to start again on the GitHub actions stuff in a separate branch. None of it should have made it to `main` in the first place, so we decided to set that aside in a separate branch (first_deployment) and move the `main` pointer back a few commits to before Shiv started work on it. In hindsight, a revert commit would have been a better way of solving that problem, so sorry for doing that.

FridaTveit:
Okay. For future reference @benjimarshall @SChandarana @kwunkiuma you should never force push to main/master. It can cause all sorts of confusion and is a good way to lose work. If necessary you can use revert commits, although even that should rarely be needed.

FridaTveit (approved):
Looks good! 🙂

PR #2, opened 2020-11-10
Add a basic interface with Google Sheets API

benjimarshall:
A demo of reading and writing from a private Google Sheet. Nice credential
storage has not been set up nicely yet, credentials and access to the private
Google Sheet can be provided if you DM me on Slack.


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+                new FileStream(path, FileMode.Open, FileAccess.Read);
+
+            var googleCredential = GoogleCredential.FromStreamAsync(stream, CancellationToken.None);
+
+            return (await googleCredential)
+                .CreateScoped(scopes)
+                .UnderlyingCredential as ServiceAccountCredential;
+        }
+
+        public EventsSheetsService() { }
```
FridaTveit, line 34: I don't think this is necessary, an empty constructor should be provided by default 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public class EventsSheetsService
+    {
+        private static readonly string[] scopes = { SheetsService.Scope.Spreadsheets };
+        private static readonly string applicationName = "Softwire Discord Bot";
+        private static readonly string spreadsheetId = "";
+
+        private SheetsService sheetsService;
+
+        private static async Task<ServiceAccountCredential> GetCredentialsAsync(string path = "credentials.json")
```
FridaTveit, line 22: Public methods before private ones please 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+            var googleCredential = GoogleCredential.FromStreamAsync(stream, CancellationToken.None);
+
+            return (await googleCredential)
+                .CreateScoped(scopes)
+                .UnderlyingCredential as ServiceAccountCredential;
+        }
+
+        public EventsSheetsService() { }
+
+        public async Task StartService()
```
FridaTveit, line 36: Is there a reason not to do this in the constructor? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            var credential = await GetCredentialsAsync();
+
+            // Create Google Sheets API service.
```
FridaTveit, line 40: Is this comment necessary? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            var range = "A2:B27";
+            var request =
+                sheetsService.Spreadsheets.Values.Get(spreadsheetId, range);
+
+            var response = await request.ExecuteAsync();
+            var values = response.Values;
+            if (values != null && values.Count > 0)
+            {
+                Console.WriteLine("Alpha, Numeric");
```
FridaTveit, line 58: Are the console logs a way to test that it's working?

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+                foreach (var row in values)
+                    Console.WriteLine($"{row[0]}, {row[1]}");
```
FridaTveit, line 60: Curly brackets please 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public async Task ReadColumns()
+        {
+            var range = "A2:B27";
+            var request =
+                sheetsService.Spreadsheets.Values.Get(spreadsheetId, range);
+
+            var response = await request.ExecuteAsync();
+            var values = response.Values;
+            if (values != null && values.Count > 0)
```
FridaTveit, line 56: I'd swap this if round and do an early return if this isn't satisfied, to avoid nesting 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                Console.WriteLine("No data found.");
+            }
+        }
+
+        public async Task WriteRow()
+        {
+            var valueRange = new ValueRange
+            {
+                Values = new List<IList<object>>
```
FridaTveit, line 72: Do you need to use `object`, is it not possible to use a more specific type? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+                        "Test write:",
+                        1
+                    }
+                },
+                Range = "A30:B30"
+            };
+            var request = sheetsService.Spreadsheets.Values.Update(valueRange, spreadsheetId, "A30:B30");
+            request.ValueInputOption = UpdateRequest.ValueInputOptionEnum.RAW;
+
+            var response = await request.ExecuteAsync();
```
FridaTveit, line 85: No need to store the response in a variable if you're not going to use it 🙂

DiscordBot.DataAccess/Program.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Threading.Tasks;
+
+namespace DiscordBot.DataAccess
+{
+    class Program
```
FridaTveit, line 5: I wouldn't expect to have a program here? As this should be used by the rest of the project?

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Threading.Tasks;
+using Google.Apis.Auth.OAuth2;
+using Google.Apis.Services;
+using Google.Apis.Sheets.v4;
+using Google.Apis.Sheets.v4.Data;
+using static Google.Apis.Sheets.v4.SpreadsheetsResource.ValuesResource;
+
+namespace DiscordBot.DataAccess
+{
+    public class EventsSheetsService
```
FridaTveit, line 14: I would expect this to have an interface as well, so that it can be used by the rest of the project 🙂

FridaTveit (commented):
Nice start! I've left a few comments, let me know if any of it doesn't make sense! 🙂

kwunkiuma (approved):
No obvious conflicts.

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: The reason I'm doing this outside the constructor is that starting the service needs to run some asynchronous code. I suppose I could just block on it because the bot can't start running until this service has been set up 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, this is just a demo of getting Google Sheets to talk to the C# code. This will all be removed and replaced eventually by more useful methods implementing methods that Frank is finishing off the interfaces for 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: `Values` is of type `IList<IList<object>>`, and I would need to cast whatever variable I am using to that type before I could assign it. An alternative might be:

``` C#
Values = new List<IList<int>>
                {
                    new List<int>
                    {
                        1,
                        1
                    }
                } as IList<IList<object>>,
```
but I'm not sure if this is any nicer 🙂

DiscordBot.DataAccess/Program.cs
benjimarshall: I was using this program to test my service out without connecting it to the rest of the Discord bot. Is there somewhere else I could put this instead, maybe in another project? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep I believe Frank (@kwunkiuma) is in the process of writing one, we were going to put that up for review once this had gone through if that is ok? We could add it now otherwise 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: Fair enough, I think a StartService method makes sense in that case 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: I see. Shame, but fair enough 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah I agree. Removed in 009105c 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, sorry, done in 009105c 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Done this in 009105c 🙂

DiscordBot.DataAccess/Program.cs
FridaTveit: If it's just used for local testing then I don't think it should be committed to the repo 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Nope, left over from a template. Sorry about that. Removed in 009105c 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Added in 009105c 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah that's a good idea. Done in 009105c 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, sorted in 009105c 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: Ah okay. For next time I would recommend creating the interface first, or creating it with the implementation of the class, so that it's clear how it will be used. I think one of these approaches would work well:
* Person A and person B discuss which methods are needed in the interface
* Person A implements the service class including the interface
* In the meantime person B uses hardcoded data instead of the service class
* Whoever merges last deals with replacing the hardcoded data with the real values from the service class

or:
* Person A and person B discuss which methods they want on the interface
* Person A or person B open a PR with the interface + a dummy implementation (returning hardcoded values)
* Person A implements the service class
* Person B uses the dummy implementation of the service class in the meantime

Does that make sense? 🙂

DiscordBot.DataAccess/Program.cs
benjimarshall: Ah ok! Removed in bd26a82 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        private static readonly string[] scopes = { SheetsService.Scope.Spreadsheets };
+        private static readonly string applicationName = "Softwire Discord Bot";
+        private static readonly string spreadsheetId = "";
```
FridaTveit, line 18: I assume this isn't meant to be empty string, is this the bit you haven't set up to be stored securely yet? If so I'd recommend you get the value from the configuration object. You can set it using [secret manager](https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-3.1&tabs=windows#secret-manager) for local development and using [environment variables](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#environment-variables) when deployed 🙂

FridaTveit (commented):
One question, otherwise looks good 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah that makes sense. I was expecting to have more trouble implementing a basic Google Sheets API than I did so the original plan was for Frank to have had that interface made before I had anything worth pushing. 

That said, maybe I should have done some more work on this before I put it up for review, waiting for the interfaces (/helping Frank write them) and implementing them all instead of just pushing the skeleton of a service with no actual features. I could help Frank with writing the interfaces and implement this as part of this PR if you like? 🙂

FridaTveit: I think putting something up for review now was the right call, but for future reference I think the interface for this class should have been agreed first as part of planning. It should just be one interface for this service as far as I understand, no need to agree any more interfaces. 

For right now I think the best thing to do is to get this merged then prioritise getting an interface up for review with mock implementations. Then you can focus on implementing that interface properly and @kwunkiuma can focus on using it. Does that make sense? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah this is the bit I haven't set up secure storage for. Shiv is trying to get it all working with the Discord secrets (as part of him getting CI/CD working), so I was planning to copy what he did for these secrets once he had got it all working, if that's ok? 🙂

SChandarana: In my local code I have used the visual studio build in environment variables thing where it adds them to a launchSettings.json and you can use `Environment.GetEnvironmentVariable()` to access it.

DiscordBot.DataAccess/EventsSheetsService.cs
kwunkiuma: There was a bit of miscommunication with this, we had agreed on a general outline of methods that the service should provide but we both assumed that the other person would include it in their PR. We've sorted out the misunderstanding now and Benji has a few methods to work on implementing while I focus more on the conversation side with the bot.

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: @benjimarshall you will only be able to use the secrets to set the environment variables as you can't access the secrets in the code. And that won't work for setting it locally. So I'd still recommend environment variables and secret manager. But fine to do in a separate PR after Shiv gets his stuff working 🙂 

@SChandarana you can use launchSettings.json but then you need to make sure you don't commit that file which is a bit annoying. So secret manager might be easier 🙂 I haven't used env variables with azure deployment before, but let me know if you're having problems with it as I'm happy to have a look 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: Sounds good! 🙂

FridaTveit (approved):
Looks good 🙂

PR #3, opened 2020-11-10
Create Workflow yml

SChandarana:
Just checking if this workflow file seems okay before pushing :)


SChandarana:
So apparently the azure workflow setup thing automatically set up the yml for me

PR #5, opened 2020-11-10
Add interface and mockup for EventsSheetsService

benjimarshall:
This is the interface and a mock-service for the Google Sheets API service that Frank and I worked through this afternoon. Here is my plan for the spreadsheet behind it all, which drove some of the decisions for the interface:

One of the sheets contains the metadata for each event, where each row refers to a different event. Currently the metadata is quite basic, but could be expanded to include things like event recurrence:

Event key | Name | Description | Time
---------- | ------- | ------------- | ----
1 | Football | 5-aside | 16/11/2020 18:00
2 | Film night | Raiders of the Lost Ark | 13/11/2020 20:00

Also, each event will have its own sheet in the spreadsheet. This will contain information about signups, where each row corresponds to a person and each column corresponds to a role in the event, so cells contain a boolean on whether a user has signed up to that role or not for the event. The plan is to have users respond to events using Discord emotes, with different emotes corresponding to different roles, so those emotes are also stored as part of this table. This per-event sheet will look something like this:

User id | Goalkeeper | Defender | Midfield | Striker | Referee | Spectator
-------- | ------------ | ---------- | --------- | ------- | --------- | ----------
Emote | :key: | :shield: | :runner: | :dart: | :clipboard: | :telescope:
Alice | 0 | 1 | 0 | 0 | 0 | 0
Bob | 0 | 0 | 0 | 1 | 0 | 0
Charlie | 0 | 0 | 1 | 0 | 0 | 0
David | 1 | 1 | 0 | 0 | 0 | 0
Eve | 0 | 0 | 0 | 0 | 1 | 0

The mock service also contains a mock event, and aims to mimic the way the real service will behave when interacting with sheets of this layout. The grid-like data storage has been implemented as nested dictionaries in the mock service, which I hope is ok 🙂


DiscordBot.DataAccess/MockEventsSheetsService.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using DiscordBot.Models;
+
+namespace DiscordBot.DataAccess
+{
+    public class MockEventsSheetsService : IEventsSheetsService
```
FridaTveit, line 9: Sorry, I probably wasn't entirely clear when I described creating a mock. I didn't mean a separate class, I meant make the real class implement the methods but with mock implementations for now. Then the only thing you'd need to change later would be to replace the mock implementations. It doesn't even need to keep track of the events or anything, just return something of the correct type 🙂

DiscordBot.DataAccess/IEventsSheetsService.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using DiscordBot.Models;
+
+namespace DiscordBot.DataAccess
+{
+    interface IEventsSheetsService
```
FridaTveit, line 8: I would expect this to live in the same file as the implementation class 🙂

DiscordBot.DataAccess/IEventsSheetsService.cs
```diff
@@ -0,0 +1,22 @@
...
+        Task EditEventAsync(int key, string? description, string? name, DateTime? time);
+        Task RemoveEventAsync(int key);
+
+        Task<DiscordEvent> GetEvent(int eventKey);
+        Task<IEnumerable<DiscordEvent>> ListEventsAsync();
+
+        Task<string> GetRoleOfEmote(int eventKey, string emote);
+
+        Task EditSignUp(int eventKey, string userId, string role, bool roleStatus);
+        Task<Dictionary<string, Dictionary<string, bool>>> GetSignUps(int key);
```
FridaTveit, line 20: A note for the future, it's a good idea to focus on a small part of the functionality first. For example just adding a new event. Then the workflow with be something like:
1. Add interface method(s) for new functionality
2. Use interface method(s) on the front end
3. Implement the interface method(s)

with 2 and 3 possibly being done in parallel. Even now that you've added all the interface methods I would recommend implementing it one small functionality at a time, and agreeing between you what you should work on first. FYI @kwunkiuma 🙂

DiscordBot.DataAccess/IEventsSheetsService.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using DiscordBot.Models;
+
+namespace DiscordBot.DataAccess
+{
+    interface IEventsSheetsService
+    {
+        Task AddEventAsync(string name, string description, DateTime time, Dictionary<string, string> roleEmotes);
```
FridaTveit, line 10: I'd recommend leaving out the roles to start with. You want to start with simple functionality, get that working then extend it 🙂

FridaTveit (commented):
A few comments, otherwise looks good! Sorry if it wasn't clear what I meant by a mock implementation earlier! 🙂

DiscordBot.DataAccess/MockEventsSheetsService.cs
benjimarshall: Ah I see! I've simplified it in 729f3d0 as you've suggested, with dummy methods added to the main `EventsSheetsService`, instead of having a fully featured mock events service 🙂

DiscordBot.DataAccess/IEventsSheetsService.cs
benjimarshall: Yep I agree, especially when there isn't a dedicated mock service. Changed in 729f3d0 🙂

DiscordBot.DataAccess/IEventsSheetsService.cs
benjimarshall: Yeah I agree, @kwunkiuma and I were a bit undecided on the whole roles/signups interface so it makes sense to leave that for now and sort it when the first part of the service is done. I've simplified the interface to just editing the events metadata sheet (the first table above) for the time being, so events can be added, edited and removed but no-one can sign up to them yet. I hope that's ok 🙂

DiscordBot.DataAccess/IEventsSheetsService.cs
benjimarshall: Ah ok, that's useful to know, thank you! 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -71,6 +81,31 @@ public async Task WriteRow()
             await request.ExecuteAsync();
         }
 
+        public async Task AddEventAsync(string name, string description, DateTime time)
+        {
+        }
+
+        public async Task EditEventAsync(int key, string? description, string? name, DateTime? time)
```
FridaTveit, line 88: Should the time really be optional? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -71,6 +81,31 @@ public async Task WriteRow()
...
+        {
+        }
+
+        public async Task RemoveEventAsync(int key)
+        {
+        }
+
+        public async Task<DiscordEvent> GetEventAsync(int eventKey)
+        {
+             return new DiscordEvent("Christmas Day", "Christmas!", eventKey, new DateTime(2020, 12, 25));
```
FridaTveit, line 98: Great event! 😄 🎄

FridaTveit (commented):
One question, otherwise looks great 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
kwunkiuma: We wanted to use the method to generalize editing data on the sheet, so the field(s?) that the user updates would be passed as parameters and the rest would be left as is.

benjimarshall: The idea behind this was to used named parameters and optional arguments, so you could use something like `EditEventAsync(eventKey, name: "NewName)`, and any arguments left as `null` are interpreted as fields that are not being updated. The only required argument would then be `key`, so the service knows what event is being updated.

I think I broke this feature when merging the services together, so I've added it back in d756304 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: 🦌 🦌 🛷 🎅

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: Ah I see! I was thinking of the date as a timestamp for the change, not the date of the event. That makes sense then thanks 🙂

FridaTveit (approved):
Looks good 🙂

PR #6, opened 2020-11-11
Add basic interactivity with bot

kwunkiuma:
The user can follow up the ?event command with the create operation. This operation initiates a conversation with the user where the bot would ask the user for various details about the event.

The conversation will terminate after 30 seconds of inactivity or if the user responds "stop".

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -67,21 +67,17 @@ namespace DiscordBot.Commands
                 return;
             }
 
-            DateTimeOffset eventTime;
-
-            while (true)
+            await context.RespondAsync($"{context.Member.Mention} - what time is your event?");
+            var eventTimeResponse = await GetUserResponse(context, interactivity);
+            if (eventTimeResponse == null)
             {
-                await context.RespondAsync($"{context.Member.Mention} - what time is your event?");
-                var eventTimeResponse = await GetUserResponse(context, interactivity);
-                if (eventTimeResponse == null)
-                {
-                    return;
-                }
+                return;
+            }
 
-                if (DateTimeOffset.TryParse(eventTimeResponse, out eventTime))
-                {
-                    break;
-                }
+            if (!DateTimeOffset.TryParse(eventTimeResponse, out var eventTime))
+            {
+                await context.RespondAsync($"{context.Member.Mention} - operation stopped. (invalid date format)");
+                return;
             }
 
             var embed = new DiscordEmbedBuilder
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -95,7 +95,9 @@ namespace DiscordBot.Commands
             await context.RespondAsync(embed: embed);
         }
 
-        private static async Task<string?> GetUserResponse(CommandContext context, InteractivityModule interactivity,
+        private static async Task<string?> GetUserResponse(
+            CommandContext context,
+            InteractivityModule interactivity,
             string[]? validStrings = null)
         {
             var response = interactivity.WaitForMessageAsync(
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -68,21 +68,9 @@ namespace DiscordBot.Commands
             }
 
             await context.RespondAsync($"{context.Member.Mention} - what time is your event?");
-            var eventTimeResponse = await GetUserResponse(context, interactivity);
-            if (eventTimeResponse == null)
+            var eventTime = await GetUserTimeResponse(context, interactivity);
+            if (eventTime == null)
             {
-                return;
-            }
-
-            DateTimeOffset eventTime;
-
-            try
-            {
-                eventTime = DateTimeOffset.Parse(eventTimeResponse);
-            }
-            catch (FormatException exception)
-            {
-                await context.RespondAsync($"{context.Member.Mention} - operation stopped: {exception.Message}");
                 return;
             }
 
@@ -119,6 +107,29 @@ namespace DiscordBot.Commands
             }
 
             return response;
+        }
+
+        private static async Task<DateTimeOffset?> GetUserTimeResponse(
+            CommandContext context,
+            InteractivityModule interactivity)
+        {
+            var response = await GetUserResponse(context, interactivity);
+
+            if (response == null)
+            {
+                return null;
+            }
+
+            try
+            {
+                return DateTimeOffset.Parse(response);
+            }
+            catch (FormatException exception)
+            {
+                await context.RespondAsync($"{context.Member.Mention} - operation stopped: {exception.Message}");
+                return null;
+            }
+
         }
 
         private static bool IsValidResponse(DiscordMessage response, CommandContext context, string[]? validStrings)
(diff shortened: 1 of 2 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -1,9 +1,10 @@
-﻿using System;
-using System.Collections.Generic;
+﻿#nullable enable
```
FridaTveit, line 1: You should be able to do this for the whole project, by adding it to the project file. Let me know if you have any issues with it 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -19,34 +20,109 @@ internal class EventCommands
         };
 
         [Command("event")]
+        [Description("Initiates the wizard for event-related actions")]
```
FridaTveit, line 23: How does this description get used? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -19,34 +20,109 @@ internal class EventCommands
...
+                    await context.RespondAsync("Not implemented yet!");
+                    break;
+                case "edit":
+                    await context.RespondAsync("Not implemented yet!");
+                    break;
+            }
+        }
 
-            if (response == null)
+        public async Task CreateEvent(CommandContext context, InteractivityModule interactivity)
```
FridaTveit, line 55: Does this need to be public? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -19,34 +20,109 @@ internal class EventCommands
...
+            var eventDescription = await GetUserResponse(context, interactivity);
+            if (eventDescription == null)
             {
-                await context.RespondAsync(response.Message.Content);
+                return;
             }
+
+            DateTimeOffset eventTime;
+
+            while (true)
```
FridaTveit, line 73: Why is just this bit in a while loop? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -19,34 +20,109 @@ internal class EventCommands
...
+                Title = eventName,
+                Description = eventDescription,
+                Color = new DiscordColor(0xFFFFFF),
+                Timestamp = eventTime
+            };
+
+            await context.RespondAsync(embed: embed);
+        }
+
+        private static async Task<string?> GetUserResponse(CommandContext context, InteractivityModule interactivity,
```
FridaTveit, line 99: Each parameter on a new line please 🙂

FridaTveit, line 99: Why does this need to be static? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -19,34 +20,109 @@ internal class EventCommands
...
+                case null:
+                    await context.RespondAsync($"{context.User.Mention} - timed out.");
+                    break;
+            }
+
+            return response;
         }
 
-        private bool IsValidResponse(string response)
+        private static bool IsValidResponse(DiscordMessage response, CommandContext context, string[]? validStrings)
```
FridaTveit, line 121: Why does this need to be static? 🙂

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
kwunkiuma: There is a built-in help command that lists each available command and the description.

FridaTveit: I see, thanks 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: That section is currently being looped until a response that can be parsed as a time is given

DiscordBot/Commands/EventCommands.cs
kwunkiuma: ReSharper gave me a squiggly line telling me to convert it to static, is there some convention for setting methods as static?

DiscordBot/Commands/EventCommands.cs
FridaTveit: I'm not sure a loop is necessary, I think just a descriptive error if the time can't be parsed should be enough 🙂

DiscordBot/Commands/EventCommands.cs
FridaTveit: Ah I was going to say it shouldn't if you access the valid strings as a private variable, but I realise now that you don't want to always pass them in. And actually that variable is static anyway so that shouldn't matter! So ignore me, feel free to leave 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -19,34 +19,107 @@ internal class EventCommands
...
+            await context.RespondAsync($"{context.Member.Mention} - what time is your event?");
+            var eventTimeResponse = await GetUserResponse(context, interactivity);
+            if (eventTimeResponse == null)
             {
-                await context.RespondAsync($"{context.Member.Mention} - timed out");
+                return;
             }
-            else
+
+            if (!DateTimeOffset.TryParse(eventTimeResponse, out var eventTime))
```
FridaTveit, line 77: I think it would be useful to get some more information about what was wrong with the date input, otherwise the user might get confused. Can we use `DateTime.Parse` (which should throw an exception if given invalid date) instead or will that return the wrong type for the DiscordEmbelBuilder? Or is there a version of the DateTimeOffset parse method which will throw an exception instead of returning a boolean? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -19,34 +19,107 @@ internal class EventCommands
...
+
+            var embed = new DiscordEmbedBuilder
+            {
+                Title = eventName,
+                Description = eventDescription,
+                Color = new DiscordColor(0xFFFFFF),
+                Timestamp = eventTime
+            };
+
+            await context.RespondAsync(embed: embed);
```
FridaTveit, line 91: What does the response look like? 🙂

FridaTveit (commented):
A couple of questions, otherwise looks good 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: image
For now the bot only responds with an embedded message with the details provided.

FridaTveit: Nice, that's a good start! 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -74,9 +74,15 @@ public async Task CreateEvent(CommandContext context, InteractivityModule intera
                 return;
             }
 
-            if (!DateTimeOffset.TryParse(eventTimeResponse, out var eventTime))
+            DateTimeOffset eventTime;
```
FridaTveit, line 77: It might be a bit neater to have a separate method to `GetUserTimeReponse` which will get the response, try to parse it, write the error to the user and return null if the parsing fails. Then here you'll just need the null check like with the other user inputs. What do you think? 🙂

FridaTveit (commented):
One suggestion, otherwise looks good 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: Sounds like a good idea, it should also many any future time queries easier.

FridaTveit (approved):
Looks great 🙂

PR #7, opened 2020-11-11
Fix comma missing on token and missing dependency

SChandarana:
Just fixing something that was missing from the main branch (most likely due to the weird force pushing on main earlier on)


FridaTveit (approved):
Looks good 🙂

PR #8, opened 2020-11-11
Add ability to add events to sheet

benjimarshall:
To do this: add tracking of largest key in use, and get this from the
sheet on service construction. Also, add a dictionary to map between
event keys and rows; this will come in useful when editing and
deleting events, and make sure the event adding method keeps these
up to date.

For context, when events are deleted either the will be blank rows in
the sheet, or the event key will not correspond to the row number of
the event. I chose the second option, so we need to keep track of this
mapping.

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using DiscordBot.Models;
+using Google;
 using Google.Apis.Auth.OAuth2;
 using Google.Apis.Services;
 using Google.Apis.Sheets.v4;
@@ -28,6 +30,8 @@ namespace DiscordBot.DataAccess
         private static readonly string spreadsheetId = "";
 
         private readonly SheetsService sheetsService;
+        private int largestKey;
+        private readonly Dictionary<int, int> keyToRowNumberMap;
 
         public EventsSheetsService()
         {
@@ -38,6 +42,37 @@ namespace DiscordBot.DataAccess
                 HttpClientInitializer = credential,
                 ApplicationName = applicationName
             });
+
+            try
+            {
+                var request = sheetsService.Spreadsheets.Values.Get(spreadsheetId, "EventsMetadata!A:A");
+                var response = request.Execute();
+
+                if (response == null || response.Values.Count < 1)
+                {
+                    throw new EventsSheetsInitialisationException("Metadata sheet is empty");
+                }
+
+                // Find which row each event is on, and the largest key in use
+                keyToRowNumberMap = new Dictionary<int, int>();
+                var keys = response.Values.Skip(1);
+                largestKey =
+                    keys
+                    .Select((keyRow, rowNumber) =>
+                        {
+                            var key = int.Parse((string) keyRow[0]);
+                            keyToRowNumberMap.Add(key, rowNumber + 1);
+                            return key;
+                        })
+                    .Max();
+            }
+            catch (GoogleApiException exception)
+            {
+                throw new EventsSheetsInitialisationException(
+                    $"Events Sheets Service couldn't initialise",
+                    exception
+                );
+            }
         }
 
         public async Task ReadColumns()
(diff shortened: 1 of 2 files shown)


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -83,6 +118,39 @@ public async Task WriteRow()
...
+                newRow,
+                spreadsheetId,
+                "A:E"
+            );
+            request.ValueInputOption = AppendRequest.ValueInputOptionEnum.RAW;
+
+            var response = await request.ExecuteAsync();
+
+            // Record which row the new event was inserted on
+            keyToRowNumberMap.Add(
```
FridaTveit, line 149: Is there any disadvantages with having empty rows in the sheet? Because having to keep track of this mapping doesn't seem ideal. Or even if we don't leave empty rows is it not possible to edit/delete rows based on key rather than row number?

And either way this doesn't seem necessary for the functionality right now which just allows inserting? Try to not future proof too much, instead focus on getting a simple neat implementation of the current functionality. That way you're more likely to end up with a simple and neat implementation in the end 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -38,6 +42,37 @@ public EventsSheetsService()
                 HttpClientInitializer = credential,
                 ApplicationName = applicationName
             });
+
+            try
```
FridaTveit, line 46: I'd extract this to a method as the constructor is getting rather large 🙂

FridaTveit (commented):

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I have also been implementing the edit event and delete event methods through the course of the day, but I only thought my add event method was ready to go up for review tonight. Editing events was what drove me to add this functionality, but that method is a bit messy at the moment and I wasn't planning to put it up for review until tomorrow morning.

If we don't keep track of row numbers (either implicitly when row numbers correspond to the key, or explicitly using a dictionary like the current implementation) then to edit a row we must first search through the sheet to find the row the event is being stored in. This didn't seem ideal to me because we could be maintaining this state every time we perform an operation, and we are limited to 10 API calls a second to Google Sheets and a search and then update would require 2 API calls instead of 1. I think this might be worse than keeping track of row numbers in some way, but I could be wrong 🙂

Leaving empty rows would get around this, because we know which row an event is on just by its key number. I started with this, and then to ensure that we could only edit events that hadn't been deleted I started keeping a set much like the keys in that dictionary. At that point I then converted it over to track event key-row mappings, but I could go back to keeping a set of valid event keys if you think it's better? 🙂 Alternatively we could just allow a deleted event to be edited (inserting data into part of a row) but I thought that might not be for the best solution.

FridaTveit: Search the update seems best to me in terms of code quality and avoiding bugs. Because if we're storing the state in the google sheet then we don't really want to also store some of it ourselves, that adds complexity and adds the possibility that they would get out of sync and cause bugs. 

Do you think we're likely to go over the 10 api calls a second limit? 🙂

benjimarshall: Ok, I'll switch over to using that then! I am also keeping track of the largest key in use (to know what key to give out to the next event being added), should I switch that over to search and update as well to keep things consistent? 🙂

I think the number of API calls will eventually become a problem, I've had another look and I think the limit will be 100 calls per 100 seconds (see [here](https://developers.google.com/sheets/api/limits)), because we only have one user interacting with the sheet (rather than the 500 calls/100s for the overall project). Depending how we implement multi-role signups, this could end up being 2 calls (search/update) per signup with users signing up for multiple things in quick succession, so an initial rush after publishing an event could overwhelm the API call quota from time to time.

However, having thought about it a little more though, I don't think we're going to be able to avoid the problem just by trimming the odd API call here and there (like I was trying to do here). I think we might need some kind of wait/retry system if we overload our quota for a given 100s to wait until we're allowed to use it again, or look at buying a bigger API quota from Google. Another approach would be to coalesce requests together (you are allowed to submit large numbers of requests in a single batch and it only counts as one call) but that could be a bit complicated to implement.

FridaTveit: The largest key seems a lot simpler to keep track of, it's just the last entered key right? So I don't really mind, I think either is fine 🙂 

Sounds like we've got a few options then. Submitting requests in batch sounds good, if we can get that working. But I think we should just go with this simple solution for now and just bear in mind that this is something we'll need to deal with at some point 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Refactored in bf55cbc 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Done this in bf55cbc 🙂

FridaTveit (approved):
Looks good 🙂

PR #10, opened 2020-11-12
Update gitignore to ignore Google API credentials

benjimarshall:


FridaTveit (approved):
Looks good 🙂 Probably worth adding a README.md file which explains how to create the credentials file and where to get the credentials from. And any other useful info a new dev would need 🙂

benjimarshall:
Yes I'll write one at some point soon!

FridaTveit:
Writing a README.md might be something it would be good for @timleachsw to do, since he's just ramped up so should know what he needed to do while ramping up? What do you both think? 🙂

timleachsw:
Yes I can definitely do that!!

PR #12, opened 2020-11-12
Add events editing

benjimarshall:
Using optional, default parameters on the function to allow for
partial updates.

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -17,7 +17,12 @@ namespace DiscordBot.DataAccess
     {
         Task AddEventAsync(string name, string description, DateTime time);
         Task EditEventAsync(int key, string? description = null, string? name = null, DateTime? time = null);
-        Task RemoveEventAsync(int key);
+        Task EditEventAsync(
+            int eventKey,
+            string? description = null,
+            string? name = null,
+            DateTime? time = null
+        );
 
         Task<DiscordEvent> GetEventAsync(int eventKey);
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
@@ -117,11 +122,44 @@ namespace DiscordBot.DataAccess
         }
 
         public async Task EditEventAsync(
-            int key,
+            int eventKey,
             string? description = null,
             string? name = null,
-            DateTime? time = null)
+            DateTime? time = null
+        )
         {
+            var rowNumber = await GetEventRowNumber(eventKey);
+
+            var data = new List<ValueRange>();
+
+            if (name != null)
+            {
+                data.Add(MakeCellUpdate($"EventsMetadata!B{rowNumber}", name));
+            }
+
+            if (description != null)
+            {
+                data.Add(MakeCellUpdate($"EventsMetadata!C{rowNumber}", description));
+            }
+
+            if (time != null)
+            {
+                data.Add(MakeCellUpdate($"EventsMetadata!D{rowNumber}", time.Value.ToString("s")));
+            }
+
+            if (!data.Any())
+            {
+                return;
+            }
+
+            var updateRequest = new BatchUpdateValuesRequest()
+            {
+                ValueInputOption = "RAW",
+                Data = data
+            };
+
+            var request = sheetsService.Spreadsheets.Values.BatchUpdate(updateRequest, spreadsheetId);
+            await request.ExecuteAsync();
         }
 
         public async Task RemoveEventAsync(int key)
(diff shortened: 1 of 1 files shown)


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -17,7 +17,12 @@ public interface IEventsSheetsService
     {
         Task AddEventAsync(string name, string description, DateTime time);
         Task EditEventAsync(int key, string? description = null, string? name = null, DateTime? time = null);
-        Task RemoveEventAsync(int key);
+        Task EditEventAsync(
```
FridaTveit, line 20: Did you mean to add a new method? Why not just use the one above? 😕

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -117,11 +122,44 @@ public async Task AddEventAsync(string name, string description, DateTime time)
...
+            DateTime? time = null
+        )
         {
+            var rowNumber = await GetEventRowNumber(eventKey);
+
+            var data = new List<ValueRange>();
+
+            if (name != null)
+            {
+                data.Add(MakeCellUpdate($"EventsMetadata!B{rowNumber}", name));
```
FridaTveit, line 137: Is it worth storing "B", "C" and "D" as well named constants making it clear they're the column names for the name, description and time respectively (as I assume they are)? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -179,5 +217,53 @@ private int GetLargestKey()
...
+                    throw new EventNotFoundException($"Event key {eventKey} not recognised");
+                }
+
+                var rowNumber = response.Values
+                    .Skip(1)
+                    .Select((rowValues, rowIndex) =>
+                        (key: int.Parse((string)rowValues[0]), index: rowIndex)
+                    )
+                    .Where(row => row.key == eventKey)
+                    .Select(row => row.index + 2)
```
FridaTveit, line 239: Why `+ 2`??

FridaTveit (commented):

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I meant to rename one the arguments (then requiring it to be split across more lines). I think this got snarled up in Git, but I've fixed it now 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yes that's a good idea! Done that 02e76dd 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah this is a little surprising so I've added some comments in 02e76dd 🙂 It's correcting for two things: 
 - The IEnumerable lists are 0 indexed, but Google Sheets index from 1  
 - Correct for skipping row 1, which was the header row

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -179,5 +229,61 @@ private int GetLargestKey()
...
+                }
+
+                var rowNumber = response.Values
+                    // Skip header row
+                    .Skip(1)
+                    // Parse the keys
+                    .Select((rowValues, rowIndex) =>
+                        (key: int.Parse((string)rowValues[0]), index: rowIndex)
+                    )
+                    .Where(row => row.key == eventKey)
```
FridaTveit, line 255: Can't you just use `Find` here? Then you can extract the row number for just the row you want 🙂

FridaTveit (commented):
One suggestion, otherwise looks good 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I can't think of a way of using `Find` to improve this. `Find` only exists on `List`s, and this starts out as an `IList`. As far as I can tell `FirstOrDefault` is the `IEnumerable` equivalent.

I don't think I can dive in with using it straight away, because it doesn't track the index of the item in the list (I would need to insert this using something like `Select((rowValues, rowIndex) => (rowValues, rowIndex))` I think). After this, if I run a `FirstOrDefault` I get `(0, 0)` if the find fails, but this also could be a valid value. I could cast it to a nullable `(int key, int index)?` here and produce this instead, which is one line shorter? 

<details>
<summary>Code</summary>

``` C#
response.Values
    .Skip(1)
    .Select((values, index) => (values, index))
    .Cast<(IList<object> values, int index)?>()
    .FirstOrDefault(row => int.Parse((string)row?.values[0]) == eventKey);
```
</details>

I'm not sure if that's made it any clearer though, so I thought I'd check 🙂 ReSharper is a bit less happy with it because of the `row?` part potentially being `null`, even though it can't be `null` because it's based on non-`null` values that has just been typecast to be nullable.

FridaTveit: Yes, I meant using `Find` instead of `Where` + `Select` + `Cast` + `First`. It's annoying that the default value is 0 though 🙁 Having to cast it all to a nullable type isn't ideal. You could use `First` instead of `FirstOrDefault`? Then you could catch the `InvalidOperationException` in the existing catch for when it doesn't find a match. What do you think? 🙂

benjimarshall: Using `First` is a better solution I think! This is what I've got instead, which drops the `Cast`, merges the `Where`, `FirstOrDefault` into `First`, and moves the `Select`'s extraction of the index and `+2` correction into the `return` statement. I've also added that catch in but haven't quoted it here. What do you think? 🙂

``` C#
var rowNumber = response.Values
    .Skip(1)  // Skip header row
    .Select((values, index) => (values, index))
    .First(row => int.Parse((string)row.values[0]) == eventKey);

// Extract row number, plus 2 to correct for two this:
// These lists are 0 indexed, but Sheets index from 1
// Correct for skipping row 1, the header
return rowNumber.index + 2;
```

FridaTveit: I like it! 😄

benjimarshall: Done in 9fb0ea6 🙂

FridaTveit (approved):
Looks great 🙂

PR #13, opened 2020-11-12
Sheets service: get events

benjimarshall:


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -177,15 +145,49 @@ public async Task RemoveEventAsync(int key)
...
+                }
+
+                return response.Values
+                    .Skip(1) // Skip header row
+                    .Select(row => new DiscordEvent(
+                        (string) row[NameColumnIndex],
+                        (string) row[DescriptionIndex],
+                        int.Parse((string) row[KeyColumnIndex]),
+                        (DateTime) row[TimeColumnIndex]
+                ));
```
FridaTveit, line 182: This line should be indented a bit more I think? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -195,7 +197,8 @@ private static ServiceAccountCredential GetCredential(string path = "credentials
 
             return GoogleCredential.FromStream(stream)
                 .CreateScoped(scopes)
-                .UnderlyingCredential as ServiceAccountCredential;
+                .UnderlyingCredential as ServiceAccountCredential
+                ?? throw new EventsSheetsInitialisationException("Credential maker returned null");
```
FridaTveit, line 201: Should be indented? 🙂

FridaTveit (commented):

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -37,11 +37,20 @@ public class EventsSheetsService : IEventsSheetsService
...
         private const string DescriptionColumn = "C";
-        private const string TimeColumn = "D";
-        private const string LocationColumn = "E";
+        private const int DescriptionIndex = 2;
+
+        private const string TimeColumnLetter = "D";
+        private const int TimeColumnIndex = 3;
+
+        private const string LocationColumnLetter = "E";
+        private const int LocationColumnIndex = 4;
```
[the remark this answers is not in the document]
benjimarshall: I've merged the pairs of `*ColumnLetter` and `*ColumnIndex` together in be4ad67 🙂 Should I also look to merge the list of these pairs into a `MetadataSheet` object? It would be nice to make that logical grouping explicit, but it would make referring to the columns a bit more verbose so I don't know which is best

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, sorry about that! Fixed in cc12dca 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, sorry about that! Fixed in cc12dca 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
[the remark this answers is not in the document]
FridaTveit: It wasn't quite what I meant (I was thinking of just a static class with all the public const strings) but I like your solution 🙂 I think it's fine to leave as it is now, it's only five variables so a lot more manageable 🙂

DiscordBot.DataAccess/SheetsColumn.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace DiscordBot.DataAccess
+{
+    internal readonly struct SheetsColumn
```
FridaTveit, line 5: I would use a class instead of a struct 🙂

FridaTveit (commented):
One tiny thing, otherwise looks good 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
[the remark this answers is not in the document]
benjimarshall: Ah I see 🙂 Well, at least it's better now

DiscordBot.DataAccess/SheetsColumn.cs
benjimarshall: Thanks for that! Changed in 405d888 🙂

FridaTveit (approved):
Looks great 🙂

PR #15, opened 2020-11-12
Add conversation for editing events

kwunkiuma:
The user can now edit existing events by interacting with the bot.

New features rely on GetEvent which is currently hardcoded and bugs may arise once that is implemented.

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -181,14 +181,13 @@ namespace DiscordBot.Commands
             {
                 var eventsSheetService = context.Dependencies.GetDependency<IEventsSheetsService>();
                 await eventsSheetService.EditEventAsync(eventKey, newDescription, newName, newTime);
-                await context.RespondAsync($"{context.Member.Mention} - changes saved!");
+                await context.RespondAsync($"{context.Member.Mention} - changes saved!", embed: eventEmbed);
             }
             catch (EventNotFoundException)
             {
                 await context.RespondAsync($"{context.Member.Mention} - operation stopped: event not found");
             }
         }
-
 
         private static async Task<DiscordEmbedBuilder?> GetEventEmbed(CommandContext context, int eventKey)
         {
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -114,78 +114,75 @@ namespace DiscordBot.Commands
                 return;
             }
 
-            await EditEventFields(context, interactivity, eventKey.Value, discordEmbed);
+            await context.RespondAsync($"{context.Member.Mention}", embed: discordEmbed);
+            await context.RespondAsync(
+                $"{context.Member.Mention} - what field do you want to edit? (``name``, ``description``, ``time``)\n" +
+                "``save`` to save changes.");
+            var editField = await GetUserResponse(context, interactivity, EventFields);
+            if (editField == null)
+            {
+                return;
+            }
+
+            await EditEventFields(context, interactivity, eventKey.Value, editField, discordEmbed);
         }
 
         private static async Task EditEventFields(
             CommandContext context,
             InteractivityModule interactivity,
             int eventKey,
+            string editField,
             DiscordEmbedBuilder discordEmbed)
         {
             string? newName = null;
             string? newDescription = null;
             DateTimeOffset? newTime = null;
 
-            while (true)
+            switch (editField)
             {
-                await context.RespondAsync($"{context.Member.Mention}", embed: discordEmbed);
-                await context.RespondAsync(
-                    $"{context.Member.Mention} - what field do you want to edit? (``name``, ``description``, ``time``)\n" +
-                    "``save`` to save changes.");
-                var editField = await GetUserResponse(context, interactivity, EventFields);
-                if (editField == null)
-                {
-                    return;
-                }
-
-                switch (editField)
-                {
-                    case "name":
-                        await context.RespondAsync($"{context.Member.Mention} - enter the new event name.");
-                        newName = await GetUserResponse(context, interactivity);
-                        if (newName == null)
-                        {
-                            return;
-                        }
-
-                        discordEmbed.Title = newName;
-                        break;
-                    case "description":
-                        await context.RespondAsync($"{context.Member.Mention} - enter the new description.");
-                        newDescription = await GetUserResponse(context, interactivity);
-                        if (newDescription == null)
-                        {
-                            return;
-                        }
-
-                        discordEmbed.Description = newDescription;
-                        break;
-                    case "time":
-                        await context.RespondAsync($"{context.Member.Mention} - enter the new event time.");
-                        newTime = await GetUserTimeResponse(context, interactivity);
-                        if (newTime == null)
-                        {
-                            return;
-                        }
-
-                        discordEmbed.Timestamp = newTime.Value.DateTime;
-                        break;
-                    case "save":
-                        try
-                        {
-                            var eventsSheetService = context.Dependencies.GetDependency<IEventsSheetsService>();
-                            await eventsSheetService.EditEventAsync(eventKey, newDescription, newName, newTime?.DateTime);
-                        }
-                        catch (EventNotFoundException)
-                        {
-                            await context.RespondAsync($"{context.Member.Mention} - operation stopped: event not found");
-                        }
-
-                        await context.RespondAsync($"{context.Member.Mention} - changes saved!");
+                case "name":
+                    await context.RespondAsync($"{context.Member.Mention} - enter the new event name.");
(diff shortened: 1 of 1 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -90,6 +99,117 @@ public async Task CreateEvent(CommandContext context, InteractivityModule intera
...
+            CommandContext context,
+            InteractivityModule interactivity,
+            int eventKey,
+            DiscordEmbedBuilder embed)
+        {
+            string? newName = null;
+            string? newDescription = null;
+            DateTimeOffset? newTime = null;
+
+            while (true)
```
kwunkiuma, line 130: I did want to wrap this loop around the call to this function from EditEvent, but would then need to add some kind of return value in order to terminate the loop from outside of this method.

Would that be preferable?

DiscordBot/Commands/EventCommands.cs
```diff
@@ -90,6 +99,117 @@ public async Task CreateEvent(CommandContext context, InteractivityModule intera
...
+
+        private static async Task EditEventFields(
+            CommandContext context,
+            InteractivityModule interactivity,
+            int eventKey,
+            DiscordEmbedBuilder embed)
+        {
+            string? newName = null;
+            string? newDescription = null;
+            DateTimeOffset? newTime = null;
```
kwunkiuma, line 128: These variables might not be needed once GetEvent is implemented since values from that can be used instead.

GetEvent currently returns hardcoded values which will overwrite the existing entry.

DiscordBot/Commands/EventCommands.cs
FridaTveit: I'm not sure a loop is needed. What about instead asking the user:
"Enter new name for event (leave blank if you want to keep current name):"
"Enter new time for event (leave blank ...)"
and so on until you have all the fields you need. What do you think? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -90,6 +99,117 @@ public async Task CreateEvent(CommandContext context, InteractivityModule intera
...
+            }
+
+            await EditEventFields(context, interactivity, eventKey.Value, embed);
+        }
+
+        private static async Task EditEventFields(
+            CommandContext context,
+            InteractivityModule interactivity,
+            int eventKey,
+            DiscordEmbedBuilder embed)
```
FridaTveit, line 124: Can you think of a more descriptive name than `embed`? 🙂

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
kwunkiuma: Discord doesn't support blank responses so that approach would require some keyword to skip each field, which I would prefer not to do.

I don't suppose recursion is an option?

FridaTveit: I don't think recursion would make the code any clearer. I'm just not sure an infinite loop is the right solution, both from a code perspective and from a user experience perspective. Because that would let them change e.g. the name again and again. If blank responses aren't supported then what about:
"Do you want to change the name, enter y/n"
if y:
"What's the new name?"
"Do you want to change the time?"
...

What do you think? 🙂

kwunkiuma: > Because that would let them change e.g. the name again and again.
I don't think this is necessarily a bad thing, since it allows the user to change their mind.
I could use while (!isFinished) then assign that value inside the loop? Would that be an improvement?

FridaTveit: They can always change their mind by just editing the event again. No, having and `isFinished` variable wouldn't really make any difference as the issue is with nesting everything thing in a loop. I think we're making the code more complicated for behaviour which isn't really necessary.

But it's your bot so you can have that behaviour if you want. In that case you need to split up this function as it's very long with several levels of nesting which makes it hard to read.

DiscordBot/Commands/EventCommands.cs
```diff
@@ -79,15 +86,123 @@ public async Task CreateEvent(CommandContext context, InteractivityModule intera
...
+
+            await EditEventField(context, interactivity, eventKey.Value, editField, discordEmbed);
+        }
+
+        private static async Task EditEventField(
+            CommandContext context,
+            InteractivityModule interactivity,
+            int eventKey,
+            string editField,
+            DiscordEmbedBuilder discordEmbed)
```
FridaTveit, line 133: This still doesn't really explain what this is. It contains the information about the event right? If so then maybe `embedEvent` or similar? Most importantly I think `event` should be in the name if that's why we're using it 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -79,15 +86,123 @@ public async Task CreateEvent(CommandContext context, InteractivityModule intera
...
+            {
+                case "name":
+                    await context.RespondAsync($"{context.Member.Mention} - enter the new event name.");
+                    newName = await GetUserResponse(context, interactivity);
+                    if (newName == null)
+                    {
+                        return;
+                    }
+
+                    discordEmbed.Title = newName;
```
FridaTveit, line 149: Since this isn't in a loop you can just call the event service from in here (possibly extract it to a new method so you don't need to duplicate code in each case statement). That way you don't need to define the new variables outside the switch and it becomes more obvious what you're doing 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -79,15 +86,123 @@ public async Task CreateEvent(CommandContext context, InteractivityModule intera
...
+            {
+                var eventsSheetService = context.Dependencies.GetDependency<IEventsSheetsService>();
+                await eventsSheetService.EditEventAsync(eventKey, newDescription, newName, newTime?.DateTime);
+            }
+            catch (EventNotFoundException)
+            {
+                await context.RespondAsync($"{context.Member.Mention} - operation stopped: event not found");
+            }
+
+            await context.RespondAsync($"{context.Member.Mention} - changes saved!", embed: discordEmbed);
```
FridaTveit, line 183: At the moment this would print event though the event wasn't found I believe, which doesn't seem right?

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
```diff
@@ -106,83 +106,90 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
-                    newTime = await GetUserTimeResponse(context, interactivity);
+                    var newTime = await GetUserTimeResponse(context, interactivity);
                     if (newTime == null)
                     {
                         return;
                     }
 
-                    discordEmbed.Timestamp = newTime.Value.DateTime;
+                    eventEmbed.Timestamp = newTime.Value.DateTime;
+                    await TryEditEvent(context, eventKey, eventEmbed, newTime: newTime?.DateTime);
```
FridaTveit, line 167: Do you really need the `?` in `newTime?.DateTime` given you've already checked for null above? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -106,83 +106,90 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
                     break;
             }
+        }
 
+        private static async Task TryEditEvent(
+            CommandContext context,
+            int eventKey,
+            DiscordEmbedBuilder eventEmbed,
```
FridaTveit, line 175: Are you using this?

DiscordBot/Commands/EventCommands.cs
```diff
@@ -106,83 +106,90 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
             }
             catch (EventNotFoundException)
             {
                 await context.RespondAsync($"{context.Member.Mention} - operation stopped: event not found");
             }
-
-            await context.RespondAsync($"{context.Member.Mention} - changes saved!", embed: discordEmbed);
         }
 
+
```
FridaTveit, line 192: Unneeded empty line 🙂

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
kwunkiuma: The compiler doesn't work if it's not included.

DiscordBot/Commands/EventCommands.cs
kwunkiuma: I forgot to include it in the response.

DiscordBot/Commands/EventCommands.cs
FridaTveit: What about `newTime.Value.DateTime`, I believe that should work?

FridaTveit (approved):
Looks good 🙂

PR #16, opened 2020-11-13
Sheets service: remove events

benjimarshall:
Extract metadata sheet name into a constant.

Add initialisation of metadata sheet id. For queries so far we have
been able to use its title, but to delete it we need to know its
`sheetId`. This does not change, and can be established whilst the
service is being constructed.

Implement remove events method.


FridaTveit (approved):
Looks good 🙂

PR #18, opened 2020-11-13
Add interface to SheetsService for events signup

benjimarshall:
@kwunkiuma and I had a look at making the next part of the interface between the `EventsSheetsService` and the Discord Bot. The table structure we settled on is:

User id | :key: | :shield: | :runner: | :dart: | :clipboard: | :telescope:
-------- | ------------ | ---------- | --------- | ------- | --------- | ----------
ResponseName | Goalkeeper | Defender | Midfield | Striker | Referee | Spectator
Alice | 0 | 1 | 0 | 0 | 0 | 0
Bob | 0 | 0 | 0 | 1 | 0 | 0
Charlie | 0 | 0 | 1 | 0 | 0 | 0
David | 1 | 1 | 0 | 0 | 0 | 0
Eve | 0 | 0 | 0 | 0 | 1 | 0

The only change is that we're indexing event responses based on emoji, rather than their names. This that makes more sense with how the discord bot will be indexing events; based on emoji reactions to a message.

The two get methods correspond to producing a dictionary indexed by the rows of this table, or the columns. 

Adding individual responses correspond to adding a reaction to the message (which the discord bot will then delete from the message when it has processed it), and clearing all responses for the user will correspond to a reaction that removes all response to an event from that user.

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -15,7 +15,12 @@ namespace DiscordBot.DataAccess
 {
     public interface IEventsSheetsService
     {
-        Task AddEventAsync(string name, string description, DateTime time);
+        Task AddEventAsync(
+            string name,
+            string description,
+            DateTime time,
+            IEnumerable<EventResponse>? responses = null
+        );
         Task EditEventAsync(
             int eventKey,
             string? description = null,
@@ -26,6 +31,12 @@ namespace DiscordBot.DataAccess
 
         Task<DiscordEvent> GetEventAsync(int eventKey);
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
+
+        Task AddResponseOfUser(int eventKey, ulong userId, string responseEmoji);
+        Task ClearResponsesOfUser(int eventKey, ulong userId);
+
+        Task<Dictionary<ulong, IEnumerable<EventResponse>>> GetSignupsByUser(int eventId);
+        Task<Dictionary<EventResponse, IEnumerable<ulong>>> GetSignupsByResponse(int eventId);
     }
 
     public class EventsSheetsService : IEventsSheetsService
@@ -59,8 +70,19 @@ namespace DiscordBot.DataAccess
             metadataSheetId = GetMetadataSheetId();
         }
 
-        public async Task AddEventAsync(string name, string description, DateTime time)
+        public async Task AddEventAsync(
+            string name,
+            string description,
+            DateTime time,
+            IEnumerable<EventResponse>? responses = null
+        )
         {
+            responses ??= new[]
+            {
+                new EventResponse(":white_check_mark:", "Yes"),
+                new EventResponse(":grey_question:", "Maybe")
+            };
+
             // Allocate new key
             largestKey++;
 
@@ -207,6 +229,34 @@ namespace DiscordBot.DataAccess
                 );
             }
         }
+
+#pragma warning disable 1998 // Disable compiler warning stating that the unimplemented functions are synchronous
+        public async Task AddResponseOfUser(int eventKey, ulong userId, string responseEmoji)
+        {
+        }
+
+        public async Task ClearResponsesOfUser(int eventKey, ulong userId)
+        {
+        }
+
+        public async Task<Dictionary<ulong, IEnumerable<EventResponse>>> GetSignupsByUser(int eventId)
+        {
+            return new Dictionary<ulong, IEnumerable<EventResponse>>
+            {
+                { 0UL, new[] { new EventResponse(":white_check_mark:", "Yes") } },
+                { 1UL, new[] { new EventResponse(":grey_question:", "Maybe") } }
+            };
+        }
+
+        public async Task<Dictionary<EventResponse, IEnumerable<ulong>>> GetSignupsByResponse(int eventId)
+        {
+            return new Dictionary<EventResponse, IEnumerable<ulong>>
+            {
+                { new EventResponse(":white_check_mark:", "Yes"), new[] { 0UL } },
+                { new EventResponse(":grey_question:", "Maybe"), new[] { 1UL } }
+            };
+        }
+#pragma warning restore 1998
 
         private static ServiceAccountCredential GetCredential(string path = "credentials.json")
         {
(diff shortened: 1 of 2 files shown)


kwunkiuma (approved):
No obvious conflicts.

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -26,6 +31,12 @@ public interface IEventsSheetsService
 
         Task<DiscordEvent> GetEventAsync(int eventKey);
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
+
+        Task AddResponseOfUser(int eventKey, ulong userId, string responseEmoji);
+        Task ClearResponsesOfUser(int eventKey, ulong userId);
```
FridaTveit, line 36: Maybe `ForUser` rather than `OfUser`? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -26,6 +31,12 @@ public interface IEventsSheetsService
...
+
+        Task<Dictionary<ulong, IEnumerable<EventResponse>>> GetSignupsByUser(int eventId);
```
FridaTveit, line 38: Why `ulong`? 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace DiscordBot.DataAccess.Models
+{
+    public class EventResponse
+    {
+        public string emoji { get; }
+        public string responseName { get; }
```
FridaTveit, line 8: Should be named starting with uppercase as they're properties 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public EventResponse(string emoji, string responseName)
+        {
+            this.emoji = emoji;
+            this.responseName = responseName;
+        }
+
+        protected bool Equals(EventResponse other)
```
FridaTveit, line 16: Do you need to override equals? 🙂

FridaTveit (commented):

DiscordBot.DataAccess/EventsSheetsService.cs
kwunkiuma: The Discord API stores and processes user IDs as `ulong`.

DiscordBot.DataAccess/Models/EventResponse.cs
benjimarshall: I think so. Without overriding this, it will use reference equality and I think value equality is what I think will be used more regularly (considering this is basically just a named pair of two strings).

I think another option would be to convert it to a `struct` instead of a class, if you would prefer that? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: I see 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
FridaTveit: Are you comparing any event responses at the moment though? 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
benjimarshall: Yes, sorry! Fixed in 1718faa 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah that's clearer 🙂 Changed in 1718faa

DiscordBot.DataAccess/Models/EventResponse.cs
benjimarshall: Not quite yet, but it will be used if anyone tries to get a specific event from the dictionary returned by `GetSignupsByResponse(...)` because this class is its key 🙂

FridaTveit: Okay, I would have probably just compared the fields in that case, but this is fine too. Fine to leave since it's been added, but in future I'd recommend not adding code before you need to use it 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            return Emoji == other.Emoji && ResponseName == other.ResponseName;
+        }
+
+        public override bool Equals(object? obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
```
FridaTveit, line 24: Is this necessary? Isn't it covered by the equals call later? 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        protected bool Equals(EventResponse other)
+        {
+            return Emoji == other.Emoji && ResponseName == other.ResponseName;
+        }
+
+        public override bool Equals(object? obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
```
FridaTveit, line 23: Would just `obj == null` work? If so, it's simpler 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
```diff
@@ -0,0 +1,34 @@
...
+            Emoji = emoji;
+            ResponseName = responseName;
+        }
+
+        protected bool Equals(EventResponse other)
+        {
+            return Emoji == other.Emoji && ResponseName == other.ResponseName;
+        }
+
+        public override bool Equals(object? obj)
```
FridaTveit, line 21: No abbreviations please 🙂

FridaTveit (commented):
A few suggestions for simplifications, otherwise looks good 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
benjimarshall: Ah this is auto-generated by ReSharper. I've tidied it up in d04752e 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
benjimarshall: Sorry this was auto-generated by ReSharper. I've tidied it up in d04752e 🙂

DiscordBot.DataAccess/Models/EventResponse.cs
benjimarshall: Yeah I think so. Changed in d04752e 🙂

FridaTveit (approved):
Looks good 🙂

PR #19, opened 2020-11-13
Add conversation for removing events

kwunkiuma:
The user can now remove events by interacting with the bot.

The bot prompts the user for an event ID and asks the user for confirmation before deleting the event.

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -15,7 +15,6 @@ namespace DiscordBot.Commands
         {
             "create",
             "remove",
-            "rm",
             "edit",
             "stop"
         };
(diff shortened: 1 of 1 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -14,6 +14,7 @@ internal class EventCommands
         private static readonly string[] EventOperations =
         {
             "create",
+            "remove",
             "rm",
```
FridaTveit, line 18: What's `rm` for? 🙂

FridaTveit (commented):
One comment, otherwise looks good 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: It was the previous keyword before renaming the command to remove, but I thought I'd leave it in there as an option.

Should I remove this?

FridaTveit: I would definitely recommend removing it if it's not used! 🙂

FridaTveit (approved):
Looks good 🙂

PR #20, opened 2020-11-13
Ignore Azure deployment settings file

benjimarshall:


.gitignore
```diff
@@ -368,3 +368,6 @@ obj/
 
 # Google APIs Credentials file
 DiscordBot.DataAccess/*credentials.json
+
+# Ignore Azure deployment settings
+DiscordBot/Settings.job
```
FridaTveit, line 373: I assume this is something which gets automatically created and therefore doesn't need to be checked in? 🙂

FridaTveit (commented):
One question, otherwise looks good 🙂

.gitignore
benjimarshall: Yeah this is autogenerated by VS when deploying to Azure from within the IDE. We don't need this file because whenever you do a new deployment this is regenerated 🙂

FridaTveit (approved):
Looks good 🙂

PR #22, opened 2020-11-13
Add conversation for showing an event

kwunkiuma:
The bot can display detailed information of a specific event specified by event key.

More information (e.g. signups) will be displayed once those features are implemented.

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -114,7 +114,7 @@ namespace DiscordBot.Commands
 
         public async Task RemoveEvent(CommandContext context, InteractivityModule interactivity)
         {
-            await context.RespondAsync($"{context.Member.Mention} - which event do you want to delete");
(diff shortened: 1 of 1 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -141,6 +146,24 @@ await context.Dependencies.GetDependency<IEventsSheetsService>()
             }
         }
 
+        public async Task ShowEvent(CommandContext context, InteractivityModule interactivity)
+        {
+            await context.RespondAsync($"{context.Member.Mention} - what is the event key?");
```
FridaTveit, line 151: Might be worth adding a little note in brackets here (and everywhere we ask for the event key), that the event key can be found in the event list? 🙂

FridaTveit (commented):
One tiny thing otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #23, opened 2020-11-16
Ignore launch settings

benjimarshall:
We are only using these files for secrets at the moment. If we set up
secret manager to do this then we should revert this commit.


FridaTveit (approved):
Looks good 🙂 

As mentioned during stand up, if there's nothing in the launchSettings which would otherwise be useful to have checked in then it's probably fine to keep storing the api keys in it, instead of needing to set up the secrets manager. FYI @SChandarana 🙂

PR #24, opened 2020-11-16
Add a message ID column to event metadata

benjimarshall:
This will allow the discord bot to convert the IDs of messages that
have been reacted to into event keys, so it can then update responses
to that event. This message ID could be null, if the message about the
event is not published straight after creating an event.

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -24,8 +24,7 @@ namespace DiscordBot.DataAccess
             int eventKey,
             string? description = null,
             string? name = null,
-            DateTime? time = null,
-            ulong? messageId = null
+            DateTime? time = null
         );
         Task AddMessageIdToEventAsync(int eventKey, ulong messageId);
         Task RemoveEventAsync(int eventKey);
@@ -100,7 +99,6 @@ namespace DiscordBot.DataAccess
                         description,
                         time.ToString("s"),
                         "Europe/London",
-                        messageId.ToString() ?? ""
                     }
                 }
             };
@@ -119,8 +117,7 @@ namespace DiscordBot.DataAccess
             int eventKey,
             string? description = null,
             string? name = null,
-            DateTime? time = null,
-            ulong? messageId = null
+            DateTime? time = null
         )
         {
             var rowNumber = await GetEventRowNumber(eventKey);
@@ -145,11 +142,6 @@ namespace DiscordBot.DataAccess
                 ));
             }
 
-            if (messageId != null)
-            {
-                data.Add(MakeCellUpdate($"{MetadataSheetName}!{MessageIdColumn}{rowNumber}", messageId));
-            }
-
             if (!data.Any())
             {
                 return;
@@ -167,7 +159,24 @@ namespace DiscordBot.DataAccess
 
         public async Task AddMessageIdToEventAsync(int eventKey, ulong messageId)
         {
-            await EditEventAsync(eventKey, messageId: messageId);
+            var rowNumber = await GetEventRowNumber(eventKey);
+
+            var cellValue = new ValueRange()
+            {
+                Values = new IList<object>[]
+                {
+                    new object[] { messageId }
+                }
+            };
+
+            var request = sheetsService.Spreadsheets.Values.Update(
+                cellValue,
+                spreadsheetId,
+                $"{MetadataSheetName}!{TimeZoneColumn.Letter}{rowNumber}"
+            );
+            request.ValueInputOption = UpdateRequest.ValueInputOptionEnum.RAW;
+
+            await request.ExecuteAsync();
         }
 
         public async Task RemoveEventAsync(int eventKey)
@@ -391,15 +400,16 @@ namespace DiscordBot.DataAccess
 
         private ulong? ParseMessageId(IList<object> row)
         {
-            // Does row contain a messageId cell?
-            // Or has it been trimmed off the end of the row because it is empty?
-            if (row.Count >= MessageIdColumn.Index + 1)
+            // The messageId might have been trimmed off the end of the row because it was empty
+            // so check for this.
+            var rowContainsMessageId = row.Count < MessageIdColumn.Index + 1;
+            if (rowContainsMessageId)
             {
                 return null;
             }
 
-            // Is the cell empty?
-            if ((string) row[MessageIdColumn.Index] != "")
+            var messageIdCellIsEmpty = (string) row[MessageIdColumn.Index] == "";
+            if (messageIdCellIsEmpty)
             {
                 return null;
             }
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -27,18 +27,18 @@ namespace DiscordBot.DataAccess
             DateTime? time = null,
             ulong? messageId = null
         );
-        Task AddMessageIdToEvent(int eventKey, ulong messageId);
+        Task AddMessageIdToEventAsync(int eventKey, ulong messageId);
         Task RemoveEventAsync(int eventKey);
 
         Task<DiscordEvent> GetEventAsync(int eventKey);
         Task<DiscordEvent> GetEventFromMessageIdAsync(ulong messageId);
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
 
-        Task AddResponseForUser(int eventKey, ulong userId, string responseEmoji);
-        Task ClearResponsesForUser(int eventKey, ulong userId);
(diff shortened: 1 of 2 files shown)


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -217,7 +228,11 @@ public async Task<IEnumerable<DiscordEvent>> ListEventsAsync()
                         (string) row[NameColumn.Index],
                         (string) row[DescriptionColumn.Index],
                         int.Parse((string) row[KeyColumn.Index]),
-                        (DateTime) row[TimeColumn.Index]
+                        (DateTime) row[TimeColumn.Index],
+                        (string) row[TimeZoneColumn.Index],
+                        row.Count >= MessageIdColumn.Index + 1 && (string) row[MessageIdColumn.Index] != ""
```
FridaTveit, line 233: Could you explain this logic please? 🙂

FridaTveit (commented):
One question otherwise looks good. How exactly will this logic be used on the front end? How does it fit in with @kwunkiuma's work? 🙂

kwunkiuma:
When a reaction is made on a message we are only able to obtain the ID for the message that is reacted to, and need to store the event key that the message refers to in order to update the correct event.

FridaTveit:
So the method to add/clear response for user will take a message id instead of an event key, is that right? The other service methods will remain the same? But if so, why does the add event and edit event methods need to change?

And which message is the id referring to exactly? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah this a bit complicated and probably should have a comment. The `messageId` might be `null`, and this conditional is deciding whether the value is `null`, or an actual ID, and parsing or returning `null` appropriately. 

To explain the conditional more specifically: the get requests trim trailing empty cells (so the length of the `row` array depends on the values in that row), so if the last cell containing this value is empty (ie the `messageId` is `null`), then `row[MessageIdColumn.Index]` will be out of bounds, so `null` should just returned instead. The second case checks for the cell being in the `row` array, but containing an empty string. This will also fail to parse, and corresponds to a `null` value. This will not happen in the current sheet layout, but if there is data in the cell to the right of the row then this cell will be `""` rather than being trimmed from the end of the array. I have explained that in a way that that makes sense? 🙂

FridaTveit: Ah okay, I think I see. It might help to extract this to a function, e.g. `ParseMessageId`, where you create suitably name variables for each part of this condition. That should hopefully make it easier to understand. What do you think? 🙂

benjimarshall:
`messageId` refers to an ID of a message in the Discord API; the message which describes an event which users will react to to sign up to the event. When this message is sent, the bot records this information in the table, using the `EditEvent` method on the sheets.

Add/clear currently takes event keys, and it will be for the Discord API to find out the event key from a message ID. I had meant to add a method `GetEventKeyFromMessageId` to allow for this, which I haven't done but will do so shortly. I could bring this all into the sheets service instead, and have it take message Ids instead of event keys, but I thought it was better to only be identifying events by their keys in all of the methods. Which do you think is better? 🙂

Add event doesn't strictly need to be changed, it can be initialised to `null` and edited afterwards. However, I added this field to it for constituency, because that every other field can be both added and edited using the `AddEvent` and `EditEvent` methods. 🙂

FridaTveit:
So the message the users can react to isn't being sent at the moment right? I mean we have no functionality on the front end to send it at the moment? Just making sure I understand how all this fits together 🙂 

I think it makes sense to identify the event by the key in most cases, because I assume the message for the user react to might not be sent straight away, although I could be wrong about that? 

If there's no use case where you'd want to set the message id when adding an event, which there isn't from what I understand, then I don't think it should be a parameter to the AddEvent method as that's a bit confusing.

I might actually be tempted to have a separate method to set the message id for an event. Because you're not editing it, you're just setting it. We're only expecting it to be called once, so I think it's worth making that difference clear 🙂

benjimarshall:
> So the message the users can react to isn't being sent at the moment right? I mean we have no functionality on the front end to send it at the moment? Just making sure I understand how all this fits together 🙂

Yes, that is one of the things @kwunkiuma will be adding as he makes the event response functionality work on the front end 🙂

> I think it makes sense to identify the event by the key in most cases, because I assume the message for the user react to might not be sent straight away, although I could be wrong about that?

Yeah that message might not be sent immediately, and I agree everything should be done just using the key 🙂

> If there's no use case where you'd want to set the message id when adding an event, which there isn't from what I understand, then I don't think it should be a parameter to the AddEvent method as that's a bit confusing.

Yeah that makes sense, I'll remove that then. If it turns out we need it later on I can always add it back in 🙂

> I might actually be tempted to have a separate method to set the message id for an event. Because you're not editing it, you're just setting it. We're only expecting it to be called once, so I think it's worth making that difference clear 🙂

Yeah that will be really simple to add and help make that clearer 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yes I agree, done in 483918a 🙂

benjimarshall:
Those changes we just discussed are now in 483918a 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -371,7 +387,24 @@ private async Task<int> GetEventRowNumber(int eventKey)
...
+            if (row.Count >= MessageIdColumn.Index + 1)
+            {
+                return null;
+            }
+
+            // Is the cell empty?
+            if ((string) row[MessageIdColumn.Index] != "")
+            {
+                return null;
+            }
```
FridaTveit, line 405: I would suggest using variable names instead of comments. For example:
```
var includesMessageIdCell = row.Count >= ...;
...
var messageIdCellValue = (string) row[MessageIdColumn.Index];
```
The `messageIdCellValue` can then also be reused when parsing, making that statement simpler 🙂 

I think you also have the first if statement the wrong way around here?

FridaTveit (commented):
| couple of suggestions otherwise looks good 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah I agree I've got the cases flipped, I'll fix that 🙂

I thought about doing this but I think I will get an `IndexOutOfRangeException` if I don't short-circuit the expression in some way, between checking the array is long enough and accessing the element is in the array. Is there a way of doing this? 🙂

FridaTveit: Yes, so I meant declare the first variable, check it, declare the second variable, check it then use it. So the first variable is only there for readability. Does that make sense? 🙂

benjimarshall: Ah I see! Done in 60650a7 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -165,6 +165,11 @@ public EventsSheetsService()
             await request.ExecuteAsync();
         }
 
+        public async Task AddMessageIdToEventAsync(int eventKey, ulong messageId)
+        {
+            await EditEventAsync(eventKey, messageId: messageId);
```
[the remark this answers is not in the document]
benjimarshall: Yeah that makes sense, I hadn't thought of that. Removed in 60650a7 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -391,15 +400,16 @@ private async Task<int> GetEventRowNumber(int eventKey)
...
+            // so check for this.
+            var rowContainsMessageId = row.Count < MessageIdColumn.Index + 1;
+            if (rowContainsMessageId)
             {
                 return null;
             }
 
-            // Is the cell empty?
-            if ((string) row[MessageIdColumn.Index] != "")
+            var messageIdCellIsEmpty = (string) row[MessageIdColumn.Index] == "";
```
FridaTveit, line 411: I would store `(string) row[MessageIdColumn.Index]` instead because then you can reuse it in the parse below, as I suggested earlier 🙂

FridaTveit (commented):
One comment, otherwise looks good 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, sorry about that! Changed in 577b877 🙂

FridaTveit (approved):
Looks good 🙂

PR #26, opened 2020-11-16
Add continuous deployment to main

SChandarana:
Continuous development worked on the test branch so i've changed the target branch in the yml to main


SChandarana:
Will this action go ahead even when it hasn't been merged with main? Since I have it to be "on push to main do this job". Just curious aha :D

FridaTveit (approved):
Looks good! Depending on how much other work there is to do you could extend this to:
1. Always run the build (but not deploy) on every branch.
2. Notify a slack channel if the build fails/succeeds. I'd recommend setting up a new slack channel for this to avoid spamming existing ones 🙂 

That can be useful so that you immediately know if something has broken the build. And if you get round to adding tests at some point it'll be even more useful as you can also track if the tests are broken. Depends how much other work there is of course, but might be worth either you or @timleachsw looking into? Especially if there's not enough other work for you all to do in parallel without getting conflicts.

If you do want to implement it then the example workflow file I sent you will probably be useful as that does something similar 🙂

FridaTveit:
@SChandarana with the current config it should only run when a change is pushed to main. So it should run when this PR is merged 🙂

SChandarana:
Yeah that sounds like a good idea for if we run out of things to do in parallel

PR #27, opened 2020-11-16
Write README.md

timleachsw:


README.md
```diff
@@ -0,0 +1,44 @@
...
+1. Have someone on the team add you to the SoftwireCambridge Discord developer team
+2. Go to https://discord.com/developers/applications and click "New Application" in the top right
+3. Name your local testing bot, making sure that the team is set to SoftwireCambridge
+![Create application](readme-pics/create-application.png)
+
+### Getting Google Sheets credentials
+
+1. Talk to someone on the team (currently Benji) who can give you Google sheets credentials, and ask for them.
+They will be given to you on [Zoho](https://vault.zoho.com/online/main), so make sure someone on the team raises a helpdesk ticket
+to add you to the "Morale Event Discord Bot" chamber
```
FridaTveit, line 19: I think you can request access yourself using HYP. So probably enough to just say to raise a HYP request to get access to the zoho chamber 🙂

README.md
```diff
@@ -0,0 +1,44 @@
...
+// Found in the json file, also from Zoho
+GOOGLE_CLIENT_EMAIL
+GOOGLE_PROJECT_ID
+GOOGLE_PRIVATE_KEY_ID
+GOOGLE_PRIVATE_KEY
+```
+2. With your bot selected in Discord for developers, go to "Bot" on the left
+![Bot settings](readme-pics/bot-settings.png)
+3. Copy the token on this page and pass it into your run configuration as `RELEASE_BOT_TOKEN`
+4. Pass in the `GOOGLE_SHEET_ID` and the four relevant fields in the JSON from Zoho
```
FridaTveit, line 40: Might be worth explaining exactly how to do this using launchSettings 🙂

FridaTveit (commented):
A few suggestions otherwise looks good! 🙂

README.md
SChandarana: Should we maybe push a template launchSettings.json which has the env variable names but no values?

FridaTveit: That wouldn't be ideal because then the launchSettings.json changes would appear in git all the time. However you could add a launchSettings.json.copy which has the env variable names but no values. Then in the readme have instructions for copying launchSettings.json.copy to launchSettings.json and setting the values. Does that make sense? 🙂

SChandarana: Yep sounds good, I'll give that a shot :D

README.md
```diff
@@ -15,8 +15,8 @@ It links up with Google Sheets to generate an at-a-glance list of all the events
 ### Getting Google Sheets credentials
 
 1. Talk to someone on the team (currently Benji) who can give you Google sheets credentials, and ask for them.
```
FridaTveit, line 17: This sentence isn't needed as you don't need anyone else to do anything to raise the hyp request 🙂

README.md
```diff
@@ -34,9 +34,13 @@ GOOGLE_PROJECT_ID
 GOOGLE_PRIVATE_KEY_ID
 GOOGLE_PRIVATE_KEY
 ```
+
+There is a template launchSettings.json.copy file in "Properties/ServiceDependencies/", all you need to do is remove the ".copy"
```
FridaTveit, line 38: You don't want to remove the copy, because you don't want to change that file. If you do then it will appear in git which will be annoying. Instead you should copy the file to launchSettings.json. Does that make sense? 🙂

FridaTveit (commented):

README.md
SChandarana: I think this was originally in there as when you are locally testing the bot, you need a separate sheet with it's own credentials and access (which benji would need to make). Do we still want that or is it better to just have one testing sheet for everyone to use?

FridaTveit: Ah I see! Is there any reason why people can't share a testing sheet? Because that would make it a bit easier for a new person to ramp up. What do you think @benjimarshall? 

If we want to keep using separate sheets then we should document here how to create a new one for a new user. Basically we want this documentation to be so good that if you were to all leave the project tomorrow, someone else could read this and know exactly what to do to keep working on the project. That's quite a high bar so we won't always achieve that, but it's what we aim for. If there's something which for some reason can't be fully documented, e.g if some permissions or something needs to be handed over manually, then we should at least mention it here and make it clear who they can talk to about it.

Does that make sense? 🙂

benjimarshall: It would work most of the time, but also probably break mysteriously if people share sheets. We're trying to use Google Sheets as an ACID database, but Sheets doesn't really offer any way of controlling "transactions" in the "database". If multiple users were trying to edit/delete events at the same time then things may behave in ways that neither user was expecting. Instead I think they would have to coordinate amongst themselves to make sure only one person was running their code with that sheet at a time. Having a sheet per person would make things a lot more straightforward 🙂

There's two general steps when it comes to making a new sheet: setting the sheet up and setting a service account up. There could be one shared service account (set of credentials and an API key basically), and it could just be added to each of these new sheets. The downside to this is that the API quota will then be stretched across multiple people's worth of tests. You could also make a new service account and add it to the sheet, which means you wouldn't be sharing an API quota, but then you would also need access to the Google Cloud project where all of these credentials are stored. There isn't a seamless way of sharing these without GSuite accounts, though I'm not aware of a Softwire GSuite accounts but they might exist. Otherwise they would need to be manually added by a manager of the Google project (which I am the only one of at the moment).

Does that all make sense? Which approach do you think would work best? 🙂

FridaTveit: I think it's fine to keep the approach you have now, as long as it's well documented and easy to hand over. For example if you're the only admin at the moment then you need to make sure you hand that over. Ideally a shared account would be used as the admin. If it's free you can probably just set it up with the team DL. If it's not free you probably need a helpdesk ticket. Does that make sense? 🙂

FridaTveit (approved):
Looks good 🙂 The credentials bit can be updated once a shared account etc has been setup. @benjimarshall are you happy handling setting that up and documenting it in a separate PR? 🙂

benjimarshall:
> Looks good 🙂 The credentials bit can be updated once a shared account etc has been setup. @benjimarshall are you happy handling setting that up and documenting it in a separate PR? 🙂

Yep, will do! 🙂

PR #28, opened 2020-11-16
Add response sheet creation to AddEventsAsync

benjimarshall:
Switch over to a single batch request to Google Sheets to:
1. Add new row to metadata table
2. Create response sheet for event
3. Populate response sheet with response columns

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -89,11 +89,11 @@ namespace DiscordBot.DataAccess
             largestKey++;
 
             var addEventMetadata =
-                SheetsServiceRequests.AddEventMetadata(metadataSheetId, largestKey, name, description, time);
+                SheetsServiceRequestsHelper.AddEventMetadata(metadataSheetId, largestKey, name, description, time);
 
-            var addResponseSheet = SheetsServiceRequests.AddResponseSheet(largestKey);
+            var addResponseSheet = SheetsServiceRequestsHelper.AddResponseSheet(largestKey);
 
-            var addResponseColumns = SheetsServiceRequests.AddResponseColumns(largestKey, responses);
+            var addResponseColumns = SheetsServiceRequestsHelper.AddResponseColumns(largestKey, responses);
 
             var requests = new BatchUpdateSpreadsheetRequest()
             {
@@ -180,7 +180,7 @@ namespace DiscordBot.DataAccess
 
             var requestParameters = new BatchUpdateSpreadsheetRequest()
             {
-                Requests = new[] { SheetsServiceRequests.RemoveEvent(metadataSheetId, rowNumber) }
+                Requests = new[] { SheetsServiceRequestsHelper.RemoveEvent(metadataSheetId, rowNumber) }
             };
 
             var request = sheetsService.Spreadsheets.BatchUpdate(requestParameters, spreadsheetId);
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -355,14 +355,6 @@ namespace DiscordBot.DataAccess
             return metadataSheet.Properties.SheetId.Value;
         }
 
-        private CellData MakeCellData(double number)
-        {
-            return new CellData()
-            {
-                UserEnteredValue = new ExtendedValue() { NumberValue = largestKey }
-            };
-        }
-
         private Request MakeAddMetadataRequest(int eventKey, string name, string description, DateTime time)
         {
             var appendCellsRequest = new AppendCellsRequest()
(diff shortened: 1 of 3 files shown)


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -84,33 +84,28 @@ public EventsSheetsService()
                 new EventResponse(":white_check_mark:", "Yes"),
                 new EventResponse(":grey_question:", "Maybe")
             };
+            var responseList = responses.ToList();
```
FridaTveit, line 87: Do you need this to be a list? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -360,6 +355,92 @@ private int GetMetadataSheetId()
             return metadataSheet.Properties.SheetId.Value;
         }
 
+        private CellData MakeCellData(double number)
```
FridaTveit, line 358: Are you using `number`?

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -360,6 +355,92 @@ private int GetMetadataSheetId()
...
+                    }
+                },
+                SheetId = largestKey,
+                Fields = "*"
+            };
+
+            return new Request() { AppendCells = addResponseColumns };
+        }
+
+        private CellData MakeCellData(string stringValue)
```
FridaTveit, line 435: I'd recommend defining this near the other method with the same name 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -360,6 +355,92 @@ private int GetMetadataSheetId()
...
+        {
+            return new CellData()
+            {
+                UserEnteredValue = new ExtendedValue() { StringValue = stringValue }
+            };
+        }
+
+
```
FridaTveit, line 443: Unneeded empty line

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -360,6 +355,92 @@ private int GetMetadataSheetId()
...
         private async Task<int> GetEventRowNumber(int eventKey)
         {
             try
```
FridaTveit, line 446: This file is quite long now. Might be worth trying to split it up, e.g. by extracting the methods to create request objects to a helper class? 🙂

FridaTveit (commented):
A few suggestions, otherwise looks good 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep! Removed in c3e7c4b 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, sorry these got split up! Moved in c3e7c4b 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I should be, thank you for spotting that! Fixed in c3e7c4b 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I pass through this list twice to make the two initial rows in the sheet (event response names and emoji). ReSharper warned me there was "multiple enumeration" happening here and recommended I do this. I can switch make to using an `IEnumerable` if that would be better? 🙂

FridaTveit: Do you need it to be a list in the make method? If not then I'd definitely recommend using `IEnumerable`. Even if you do I'd be tempted to convert to list in the make method instead of here 🙂

benjimarshall: The multiple enumeration warning happened in the request-making method, but I brought that change up to here. I've now moved the `ToList()` conversion back to the request-making method, so it now takes an `IEnumerable` as an argument 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah I agree. I've extracted those request making methods in db6b403 🙂

DiscordBot.DataAccess/SheetsServiceRequests.cs
```diff
@@ -0,0 +1,121 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using DiscordBot.DataAccess.Models;
+using Google.Apis.Sheets.v4.Data;
+
+namespace DiscordBot.DataAccess
+{
+    internal static class SheetsServiceRequests
```
FridaTveit, line 9: Maybe `SheetsServiceRequestsHelper` makes it a bit more clear what this is for? 🙂

FridaTveit (commented):
One tiny thing otherwise looks good 🙂

DiscordBot.DataAccess/SheetsServiceRequests.cs
benjimarshall: Yeah that's a better name 🙂 Renamed in 284f577

FridaTveit (approved):
Looks good 🙂

PR #29, opened 2020-11-16
Add signup message creation

kwunkiuma:
The bot can now create a signup message for members to react to.

For now, reacting to the message doesn't do anything and the message is sent to the channel where the command is called.
image

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -251,45 +251,45 @@ namespace DiscordBot.Commands
             var eventsSheetsService = context.Dependencies.GetDependency<IEventsSheetsService>();
             var discordEvent = await eventsSheetsService.GetEventAsync(eventKey);
 
-            var optionsList = await eventsSheetsService.GetSignupsByResponseAsync(eventKey);
+            var signupsByResponse = await eventsSheetsService.GetSignupsByResponseAsync(eventKey);
 
-            var signupEmbed = GetSignupEmbed(discordEvent, optionsList);
+            var signupEmbed = GetSignupEmbed(discordEvent, signupsByResponse);
 
             var signupMessage = await context.RespondAsync($"Signups are open for __**{discordEvent.Name}**__!", embed: signupEmbed);
             await eventsSheetsService.AddMessageIdToEventAsync(eventKey, signupMessage.Id);
 
-            foreach (var response in optionsList.Keys)
+            foreach (var response in signupsByResponse.Keys)
             {
                 await signupMessage.CreateReactionAsync(DiscordEmoji.FromName(context.Client, response.Emoji));
             }
         }
 
         private static Dictionary<ulong, IEnumerable<EventResponse>> GetResponsesByUser(
-            Dictionary<EventResponse, IEnumerable<ulong>> optionsList)
+            Dictionary<EventResponse, IEnumerable<ulong>> signupsByResponse)
         {
-            var userResponseDictionary = new Dictionary<ulong, IEnumerable<EventResponse>>();
+            var signupsByUser = new Dictionary<ulong, IEnumerable<EventResponse>>();
 
-            var userIds = optionsList.Values
-                .SelectMany(x => x)
+            var userIds = signupsByResponse.Values
+                .SelectMany(responseSignups => responseSignups)
                 .Distinct();
 
             foreach (var userId in userIds)
             {
-                var userResponses = optionsList
-                    .Where(x => x.Value.Contains(userId))
-                    .Select(x => x.Key);
+                var userResponses = signupsByResponse
+                    .Where(responseSignups => responseSignups.Value.Contains(userId))
+                    .Select(responseSignups => responseSignups.Key);
 
-                userResponseDictionary.Add(userId, userResponses);
+                signupsByUser.Add(userId, userResponses);
             }
 
-            return userResponseDictionary;
+            return signupsByUser;
         }
 
         private static DiscordEmbedBuilder GetSignupEmbed(
             DiscordEvent discordEvent,
-            Dictionary<EventResponse, IEnumerable<ulong>> optionsList)
+            Dictionary<EventResponse, IEnumerable<ulong>> signupsByResponse)
         {
-            var userResponseDictionary = GetResponsesByUser(optionsList);
+            var userResponseDictionary = GetResponsesByUser(signupsByResponse);
 
             var signupEmbed = new DiscordEmbedBuilder
             {
@@ -317,7 +317,7 @@ namespace DiscordBot.Commands
 
             var optionsField = string.Join(
                 "\n",
-                optionsList.Select(response => $"{response.Key.Emoji} - {response.Key.ResponseName}")
+                signupsByResponse.Select(response => $"{response.Key.Emoji} - {response.Key.ResponseName}")
             );
 
             signupEmbed.AddField("Response options", optionsField);
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -261,19 +261,26 @@ namespace DiscordBot.Commands
 
             var userResponseDictionary = await eventsSheetsService.GetSignupsByUserAsync(eventKey);
 
-            var usersField = string.Join("\n", userResponseDictionary.Select(
-                response => $"<@{response.Key}>"));
+            var usersField = string.Join(
+                "\n",
+                userResponseDictionary.Select(response => $"<@{response.Key}>")
+            );
+
             signupEmbed.AddField("Participants", usersField, true);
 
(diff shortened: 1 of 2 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -217,6 +219,72 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
+                Footer = new DiscordEmbedBuilder.EmbedFooter
+                {
+                    Text = $"event key: {eventKey}"
+                }
+            };
+
+            var userResponseDictionary = await eventsSheetsService.GetSignupsByUserAsync(eventKey);
+
+            var usersField = string.Join("\n", userResponseDictionary.Select(
+                response => $"<@{response.Key}>"));
```
FridaTveit, line 265: If you're going to split this over several lines then each argument to the function needs to be on a new line, otherwise it's a bit tricky to read. For example:
```
var usersField = string.Join(
  "\n",
  userResponseDictionary.Select(response => $"<@{response.Key}>")
);
```
or
```
var usersField = string.Join(
  "\n", 
  userResponseDictionary.Select(
    response => $"<@{response.Key}>"
  )
);
```
Does that make sense? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -217,6 +219,72 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
+
+            var responsesList = userResponseDictionary.Select(
+                response => string.Join(" ", response.Value));
+            var responsesField = string.Join("\n", responsesList);
+            signupEmbed.AddField("Responses", responsesField, true);
+
+            var optionsList = await eventsSheetsService.GetSignupsByResponseAsync(eventKey);
+            var optionsField = string.Join(
+                "\n", optionsList.Select(response =>
+                    $"{response.Key.Emoji} - {response.Key.ResponseName}"));
```
FridaTveit, line 276: I think it's more readable with the closing brackets on a new line, especially in cases like this 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -217,6 +219,72 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
+            signupEmbed.AddField("Response options", optionsField);
+
+            var signupMessage = await context.RespondAsync($"Signups are open for __**{discordEvent.Name}**__!", embed: signupEmbed);
```
FridaTveit, line 279: Should this message really be sent in the same channel as everything else? Isn't that an admin channel? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -217,6 +219,72 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
+            var confirmationResponse = await GetUserConfirmation(context, interactivity);
+            if (confirmationResponse == null || confirmationResponse == false)
+            {
+                return;
+            }
+
+            await SendSignupMessage(context, eventKey.Value);
+        }
+
+        private static async Task SendSignupMessage(CommandContext context, int eventKey)
```
FridaTveit, line 247: This method is pretty long. Can you split it up a bit? 🙂

FridaTveit (commented):
A few suggestions otherwise looks good 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: The idea was to prioritize getting functional signups so we've sent the message in the same channel for now, should I make the change now?

DiscordBot/Commands/EventCommands.cs
kwunkiuma: I assume you're referring to `SendSignupMessage()`?

DiscordBot/Commands/EventCommands.cs
FridaTveit: Ah okay, as long as you've got a definite plan to move it to a different channel then that's fine 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: I'm having some troubles doing this, can I have some suggestions?

FridaTveit: So I think splitting out creating the sign up embed would be useful. Depending how big that split out method turns out to be, you could even split that up further so that you have separate methods for adding participants to it, adding responses to it and adding response options. Does that make sense? 🙂

kwunkiuma: I'm finding it hard to do that because adding the reactions (which needs to be done after the message is sent) requires information from the optionsList variable (which is also used for the embed). Should I include that as an output parameter as well as returning the embed?

FridaTveit: In that case I would get the userResponseDictionary and optionsList here in this method. Then you can pass them in to the method to create the embed. I would not recommend using an output parameter as code using an output parameter can be a bit confusing 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -217,6 +221,89 @@ public async Task EditEvent(CommandContext context, InteractivityModule interact
...
+        { 
+            var eventsSheetsService = context.Dependencies.GetDependency<IEventsSheetsService>();
+            var discordEvent = await eventsSheetsService.GetEventAsync(eventKey);
+
+            var userResponseDictionary = await eventsSheetsService.GetSignupsByUserAsync(eventKey);
+            var optionsList = await eventsSheetsService.GetSignupsByResponseAsync(eventKey);
```
FridaTveit, line 255: Actually do you need to get both of these? They both contain all the responses and all the users right? I don't think it's worth doing two requests to get the same information? 🙂

FridaTveit (commented):
One question otherwise looks good. Nice refactor 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: Only using userResponseDictionary won't because response types with no signups won't be shown.

Only using optionsList would could work but the structure isn't ideal for displaying users alongside their respective responses, meaning I'd have to create something like the userResponseDictionary from optionsList.

FridaTveit: I don't think we should do both requests though, that seems wasteful. Because it would mean doing two google sheets requests I believe, correct me if I'm wrong @benjimarshall. If the returned structure isn't ideal then maybe the service api needs tweaking?

DiscordBot/Commands/EventCommands.cs
```diff
@@ -251,10 +251,9 @@ private static async Task SendSignupMessage(CommandContext context, int eventKey
             var eventsSheetsService = context.Dependencies.GetDependency<IEventsSheetsService>();
             var discordEvent = await eventsSheetsService.GetEventAsync(eventKey);
 
-            var userResponseDictionary = await eventsSheetsService.GetSignupsByUserAsync(eventKey);
             var optionsList = await eventsSheetsService.GetSignupsByResponseAsync(eventKey);
 
-            var signupEmbed = GetSignupEmbed(discordEvent, userResponseDictionary, optionsList);
+            var signupEmbed = GetSignupEmbed(discordEvent, optionsList);
```
FridaTveit, line 256: I'm not sure `optionsList` is the most descriptive name. What about `signupsByResponse` which matches the service method name? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -265,11 +264,33 @@ private static async Task SendSignupMessage(CommandContext context, int eventKey
             }
         }
 
+        private static Dictionary<ulong, IEnumerable<EventResponse>> GetResponsesByUser(
+            Dictionary<EventResponse, IEnumerable<ulong>> optionsList)
+        {
+            var userResponseDictionary = new Dictionary<ulong, IEnumerable<EventResponse>>();
+
+            var userIds = optionsList.Values
+                .SelectMany(x => x)
```
FridaTveit, line 273: More descriptive name? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -265,11 +264,33 @@ private static async Task SendSignupMessage(CommandContext context, int eventKey
...
+                .Distinct();
+
+            foreach (var userId in userIds)
+            {
+                var userResponses = optionsList
+                    .Where(x => x.Value.Contains(userId))
+                    .Select(x => x.Key);
```
FridaTveit, line 280: More descriptive name? 🙂

FridaTveit (commented):
A few comments on naming, otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #31, opened 2020-11-16
Remove response sheet on event removal

benjimarshall:
Also generalise the `sheetId` lookup (where sheet in this context means a single sheet among multiple in a spreadsheet document).


FridaTveit (approved):
Looks good 🙂

PR #32, opened 2020-11-17
Add responses for user

benjimarshall:
- Add method to get potential responses to an event
 - Move SheetsService exceptions to own folder
 - Tidy up and generalise finding rows in sheets
 - Implement adding responses to an event

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -32,6 +32,7 @@ namespace DiscordBot.DataAccess
         Task<DiscordEvent> GetEventAsync(int eventKey);
         Task<DiscordEvent> GetEventFromMessageIdAsync(ulong messageId);
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
+        Task<IEnumerable<EventResponse>> GetEventResponseOptionsAsync(int eventKey);
 
         Task AddResponseForUserAsync(int eventKey, ulong userId, string responseEmoji);
         Task ClearResponsesForUserAsync(int eventKey, ulong userId);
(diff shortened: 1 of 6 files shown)


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -32,6 +33,7 @@ public interface IEventsSheetsService
         Task<DiscordEvent> GetEventAsync(int eventKey);
         Task<DiscordEvent> GetEventFromMessageIdAsync(ulong messageId);
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
+        Task<IEnumerable<EventResponse>> GetEventResponseOptionsAsync(int eventKey);
```
FridaTveit, line 36: When will this be used? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -254,11 +258,104 @@ public async Task<IEnumerable<DiscordEvent>> ListEventsAsync()
...
+            {
+                throw new EventInitialisationException(
+                    $"Could not find event responses for event {eventKey}. Has it been published yet?"
+                );
+            }
+        }
+
         public async Task AddResponseForUserAsync(int eventKey, ulong userId, string responseEmoji)
         {
+            var responseRowTask = GetResponseRowNumber(eventKey, userId);
```
FridaTveit, line 295: Is this also async? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -254,11 +258,104 @@ public async Task<IEnumerable<DiscordEvent>> ListEventsAsync()
...
+            var responseColumnsTask = GetEventResponseOptionsAsync(eventKey);
+
+            await Task.WhenAll(responseRowTask, responseColumnsTask);
+            var responseRow = await responseRowTask;
+            var responseColumns = await responseColumnsTask;
```
FridaTveit, line 300: Do you need to await these again given that you've just awaited them both in your `WhenAll` statement above? 🙂

FridaTveit (commented):

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I need to unwrap them in some way. I could use `*Task.Result` instead maybe? What do you think? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I'm currently using it internally, and it could be useful in the rest of the bot but it isn't in use at the moment. I've changed it to be a private method in 96305e0 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep! Fixed in 96305e0 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
FridaTveit: Using `.Result` sounds sensible to me 🙂

benjimarshall: Changed in c520c27 🙂

FridaTveit (approved):
Looks good 🙂

PR #33, opened 2020-11-17
Migrate from 3.x to 4.x

kwunkiuma:
Turns out 3.x is almost obsolete so we've migrated the bot to the 4.x version.

Main changes:

- The service can now be assigned in the constructor and therefore doesn't need to be gotten every time it's needed
- Some classes were renamed
- Some class methods were renamed


FridaTveit (approved):
Looks good 🙂

PR #34, opened 2020-11-17
Bugfix for creating events

benjimarshall:
Stop Google Sheets trying to parse the event ID as a `double` (and then
dropping precision) by writing to the sheet a `string` version of the 
ID to the sheet.


FridaTveit (approved):
Looks good 🙂

PR #36, opened 2020-11-17
Fix the DiscordBot Build

benjimarshall:
I didn't realise renaming exceptions would break this file... Whoops!
This should import them.


FridaTveit (approved):
Looks good 🙂

PR #37, opened 2020-11-17
List responses

benjimarshall:
Get responses for an event and parse into dictionaries

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -594,7 +594,7 @@ namespace DiscordBot.DataAccess
             var newRow = responseColumns
                 .Select(response => response.Emoji == responseEmoji ? 1 : 0)
                 .Cast<object>()
-                .Prepend(userId);
+                .Prepend(userId.ToString());
 
             var values = new ValueRange()
             {
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -297,25 +297,83 @@ namespace DiscordBot.DataAccess
             await request.ExecuteAsync();
         }
 
-#pragma warning disable 1998 // Disable compiler warning stating that the unimplemented functions are synchronous
         public async Task<Dictionary<ulong, IEnumerable<EventResponse>>> GetSignupsByUserAsync(int eventId)
         {
-            return new Dictionary<ulong, IEnumerable<EventResponse>>
+            var request = sheetsService.Spreadsheets.Values.Get(
+                spreadsheetId,
+                $"{eventId}"
+            );
+            request.ValueRenderOption = GetRequest.ValueRenderOptionEnum.FORMATTEDVALUE;
+            var response = await request.ExecuteAsync();
+
+            if (response == null || response.Values.Count < 2)
             {
-                { 0UL, new[] { new EventResponse(":white_check_mark:", "Yes") } },
-                { 1UL, new[] { new EventResponse(":grey_question:", "Maybe") } }
-            };
+                throw new EventInitialisationException("Sign up sheet is empty");
+            }
+
+            var responseColumns = ParseResponseHeaders(response.Values);
+
+            return response.Values
+                .Skip(2)
+                .Select(row =>
+                {
+                    var user = ulong.Parse((string) row[0]);
+                    var responseList =
+                        row.Skip(1) // Skip title column
+                            .Zip(responseColumns)
+                            .Where<(object cell, EventResponse response)>(
+                                pair => pair.cell.ToString() == "1"
+                            )
+                            .Select(pair => pair.response);
+                    return (user, responseList);
+                })
+                .ToDictionary(
+                    entry => entry.user,
+                    entry => entry.responseList
+                );
         }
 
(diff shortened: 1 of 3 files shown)


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -297,25 +297,83 @@ public async Task ClearResponsesForUserAsync(int eventKey, ulong userId)
             await request.ExecuteAsync();
         }
 
-#pragma warning disable 1998 // Disable compiler warning stating that the unimplemented functions are synchronous
         public async Task<Dictionary<ulong, IEnumerable<EventResponse>>> GetSignupsByUserAsync(int eventId)
```
FridaTveit, line 300: I thought we concluded that it wasn't necessary to have both this and GetSignupsByResponse? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -297,25 +297,83 @@ public async Task ClearResponsesForUserAsync(int eventKey, ulong userId)
...
-            };
+                throw new EventInitialisationException("Sign up sheet is empty");
+            }
+
+            var responseColumns = ParseResponseHeaders(response.Values);
+
+            return response.Values
+                .Skip(2)
+                .Select(row =>
+                {
```
FridaTveit, line 319: I'd recommend extracting this to its own method 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -297,25 +297,83 @@ public async Task ClearResponsesForUserAsync(int eventKey, ulong userId)
...
+
+            var responseColumns = ParseResponseHeaders(response.Values).ToList();
+
+            var result = responseColumns.ToDictionary(
+                eventResponse => eventResponse,
+                eventResponse => new List<ulong>()
+            );
+
+            response.Values.Skip(2)
+                .ToList().
```
FridaTveit, line 358: Full stop should be on the next line

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -297,25 +297,83 @@ public async Task ClearResponsesForUserAsync(int eventKey, ulong userId)
...
+                ForEach(row =>
+            {
+                var user = ulong.Parse((string)row[0]);
```
FridaTveit, line 361: I'd recommend extracting this 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -576,5 +629,15 @@ IEnumerable<EventResponse> responseColumns
 
             await request.ExecuteAsync();
         }
+
+        private IEnumerable<EventResponse> ParseResponseHeaders(IList<IList<object>> sheetsResponse)
```
FridaTveit, line 633: This file is getting very long again. Might be worth extracting all the parsing to a parser class? 🙂

FridaTveit (commented):

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, I had forgotten about that! Removed in 6877d46 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Removed in 6877d46 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yep, sorry! Fixed in 73aa63b 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah I agree. Refactored in 73aa63b 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah the class is quite large. I've extracted the parsing methods as you've suggested in 73aa63b, but it's still quite large. Nearly all of the remaining methods use some of the `EventsSheetsService` state, usually the underlying Google Sheets service class. 

I could try to group these methods together into classes, but they would either need to take most of this `EventsSheetsService`'s fields as parameters, or contain the `EventSheetService` state inside their own classes. Is this a good idea? Or should I just stick with this large class file 🙂

FridaTveit: Hm, good question. I think extracting the parsing and request creation is a good start because those are completely separate concerns that it's good to split out. So just make sure that any logic to do with creating requests or parsing responses is moved out to those classes. Otherwise I think it's probably fine to keep the rest in this class at least for now. 

Just keep your eye out for anything else which can be split out as you're working. I don't think needing to pass the underlying service class or any other state as parameters is necessarily the end of the world, but it's important that the new class is genuinely separate to this service class. So that it has a separate purpose and isn't just an extension of this class, if that makes sense 🙂

benjimarshall: Yeah, will do! Thanks for that 🙂

PR #38, opened 2020-11-17
Add group command for create and fix migration issues

SChandarana:
I've converted the create command into a group command (which can utilise overloads). This means that the following things work:
?event
?event create
?event create "[name]"
?event create "[name]" "[description]"
?event create "[name]" "[description]" "[Valid DateTime]"

If any of these are missing something it will use interactivity to prompt the user to give it to them.

Whilst implementing this I removed some redundant code (the passing of the variable interactivity) so this PR does have quite a lot of small changes repeated throughout. The important changes are anything to do with group commands and the create command. The final small change was in program.cs where I needed to add intents (a 4.x thing) to give the bot permissions to read user roles (for the require roles attribute)

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -8,11 +8,12 @@ using DiscordBot.DataAccess.Models;
 using DSharpPlus.CommandsNext;
 using DSharpPlus.CommandsNext.Attributes;
 using DSharpPlus.Entities;
-using DSharpPlus.Interactivity;
 using DSharpPlus.Interactivity.Extensions;
 
 namespace DiscordBot.Commands
 {
+    [Group("event")]
+    [RequireRoles(RoleCheckMode.All, "Bot Whisperer")]
     internal class EventCommands : BaseCommandModule
     {
         private static readonly string[] EventOperations =
@@ -45,12 +46,10 @@ namespace DiscordBot.Commands
             this.eventsSheetsService = eventsSheetsService;
         }
 
-        [Command("event")]
+        [GroupCommand]
         [Description("Initiates the wizard for event-related actions")]
         public async Task Event(CommandContext context)
         {
-            var interactivity = context.Client.GetInteractivity();
-
             await context.RespondAsync(
                 $"{context.Member.Mention} - choose one of the actions below or answer ``stop`` to cancel. (time out in 30s)\n" +
                 "``create`` - create new event.\n" +
@@ -61,7 +60,7 @@ namespace DiscordBot.Commands
                 "``start`` - open signups for event."
             );
 
-            var eventOperation = await GetUserResponse(context, interactivity, EventOperations);
+            var eventOperation = await GetUserResponse(context, EventOperations);
             if (eventOperation == null)
             {
                 return;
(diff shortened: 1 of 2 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -8,11 +8,12 @@
 using DSharpPlus.CommandsNext;
 using DSharpPlus.CommandsNext.Attributes;
 using DSharpPlus.Entities;
-using DSharpPlus.Interactivity;
 using DSharpPlus.Interactivity.Extensions;
 
 namespace DiscordBot.Commands
 {
+    [Group("event")]
+    [RequireRoles(RoleCheckMode.All, "Bot Whisperer")]
```
FridaTveit, line 16: What's this for? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -70,50 +69,69 @@ public async Task Event(CommandContext context)
...
                 case "start":
-                    await CreateSignupSheet(context, interactivity);
+                    await CreateSignupSheet(context);
                     break;
             }
         }
 
-        public async Task CreateEvent(CommandContext context, InteractivityExtension interactivity)
+        [Command("create")]
+        public async Task CreateEventName(CommandContext context)
```
FridaTveit, line 93: Why the rename? It's not just creating the event name, it's creating the whole event...

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
SChandarana: So the [Group] attribute makes it so that every command in this class can be invoked with ?event [commandName].
The RequireRoles attribute checks if the user that invoked the command has all of the speicified roles (meaning people without it can't use the commands)

FridaTveit: Ah okay. And is "Bot Whisperer" the role you want to go with? Has that been configured/do you have a plan for configuring it so that only the relevant people have that role? 🙂

DiscordBot/Commands/EventCommands.cs
SChandarana: This is very true, should I change the names of all of them back to just "CreateEvent" as they all end up doing the same thing (using overloads)

FridaTveit: Yes, I think that sounds sensible 🙂

DiscordBot/Commands/EventCommands.cs
SChandarana: Currently that's the name I've picked and is what I'm using in the test server. This can easily be changed after asking the softwire discord moderators which they prefer. The plan is to have the moderators just assign the role to the dedicated event managers to give them access to the commands

FridaTveit: Okay, sounds fine. Probably worth documenting this somewhere though, for example in the readme? 🙂

SChandarana: Will do :D

SChandarana:
I'll add the things to the readme within the readme branch as it hasn't yet been merged into main

FridaTveit (approved):
Looks good 🙂

PR #39, opened 2020-11-18
Add check a response's validity before adding it

benjimarshall:
Thanks to @kwunkiuma for spotting this bug!

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -271,7 +271,7 @@ namespace DiscordBot.DataAccess
             var responseRow = responseRowTask.Result;
             var responseColumns = responseColumnsTask.Result;
 
-            if (responseColumns.All(response => response.Emoji != responseEmoji))
+            if (!responseColumns.Any(response => response.Emoji == responseEmoji))
             {
                 throw new ResponseNotFoundException(
                     $"Response {responseEmoji} is not recognised for event {eventKey}"


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -520,7 +520,16 @@ private async Task<IEnumerable<EventResponse>> GetEventResponseOptionsAsync(int
             IEnumerable<EventResponse> responseColumns
         )
         {
-            var newRow = responseColumns
+            var responseList = responseColumns.ToList();
+
+            if (responseList.All(response => response.Emoji != responseEmoji))
```
FridaTveit, line 525: Minor but I'd probably find `if (!responseList.Any(...))` a bit easier to understand? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -548,6 +557,13 @@ IEnumerable<EventResponse> responseColumns
         {
             var index = responseColumns.ToList().FindIndex(responseOption => responseOption.Emoji == responseEmoji);
 
+            if (index == -1)
+            {
+                throw new ResponseNotFoundException(
+                    $"Response {responseEmoji} is not recognised for event {eventKey}"
+                );
+            }
```
FridaTveit, line 565: Does this check really need to be duplicated in both add response methods? It looks like the logic is really the same for both so could be done for both of them in the calling method? What do you think? 🙂

FridaTveit (commented):

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah sure! Changed in 1dfe1d6 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah I agree! Changed in 32ae737 🙂

FridaTveit (approved):
Looks good 🙂

PR #40, opened 2020-11-18
Add reaction handling

kwunkiuma:
The user can now react to the signup sheet and the following will happen:

1. The Google Sheet will be updated
2. Send the user a confirmation via DM
3. Update the signup sheet
4. Remove the user's reaction

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -1,14 +1,13 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using DiscordBot.DataAccess;
 using DiscordBot.DataAccess.Exceptions;
-using DiscordBot.DataAccess.Models;
 using DSharpPlus.CommandsNext;
 using DSharpPlus.CommandsNext.Attributes;
 using DSharpPlus.Entities;
 using DSharpPlus.Interactivity.Extensions;
+using static DiscordBot.Commands.EventHelper;
 
 namespace DiscordBot.Commands
 {
@@ -287,67 +286,8 @@ namespace DiscordBot.Commands
             {
                 await signupMessage.CreateReactionAsync(DiscordEmoji.FromName(context.Client, response.Emoji));
             }
-        }
 
-        private Dictionary<ulong, IEnumerable<EventResponse>> GetResponsesByUser(
-            Dictionary<EventResponse, IEnumerable<ulong>> signupsByResponse)
-        {
-            var signupsByUser = new Dictionary<ulong, IEnumerable<EventResponse>>();
-
-            var userIds = signupsByResponse.Values
-                .SelectMany(responseSignups => responseSignups)
-                .Distinct();
-
-            foreach (var userId in userIds)
-            {
-                var userResponses = signupsByResponse
-                    .Where(responseSignups => responseSignups.Value.Contains(userId))
-                    .Select(responseSignups => responseSignups.Key);
-
-                signupsByUser.Add(userId, userResponses);
-            }
-
-            return signupsByUser;
-        }
-
-        private DiscordEmbedBuilder GetSignupEmbed(
-            DiscordEvent discordEvent,
-            Dictionary<EventResponse, IEnumerable<ulong>> signupsByResponse)
-        {
-            var userResponseDictionary = GetResponsesByUser(signupsByResponse);
-
-            var signupEmbed = new DiscordEmbedBuilder
-            {
-                Title = $"{discordEvent.Name} - {discordEvent.Time:ddd dd MMM yyyy @ h:mm tt}",
-                Description = discordEvent.Description,
-                Footer = new DiscordEmbedBuilder.EmbedFooter
-                {
-                    Text = $"event key: {discordEvent.Key}"
-                }
-            };
-
-            var usersField = string.Join(
-                "\n",
-                userResponseDictionary.Select(response => $"<@{response.Key}>")
-            );
-
-            signupEmbed.AddField("Participants", usersField, true);
-
-            var responsesList = userResponseDictionary.Select(
-                response => string.Join(" ", response.Value)
-            );
-            var responsesField = string.Join("\n", responsesList);
-
-            signupEmbed.AddField("Response(s)", responsesField, true);
-
-            var optionsField = string.Join(
-                "\n",
-                signupsByResponse.Select(response => $"{response.Key.Emoji} - {response.Key.ResponseName}")
-            );
-
-            signupEmbed.AddField("Response options", optionsField);
-
-            return signupEmbed;
+            await signupMessage.CreateReactionAsync(DiscordEmoji.FromName(context.Client, ":no_entry_sign:"));
         }
 
         private async Task EditEventField(
@@ -464,7 +404,7 @@ namespace DiscordBot.Commands
                 message =>
                     IsValidResponse(message, context, ConfirmationResponses),
                 TimeSpan.FromSeconds(30)
-            ).Result.Result.Content;
+            ).Result.Result?.Content;
 
             switch (response)
             {
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -39,12 +39,10 @@ namespace DiscordBot.Commands
         };
 
         private readonly IEventsSheetsService eventsSheetsService;
-        private readonly EventHelper eventHelper;
 
-        public EventCommands(IEventsSheetsService eventsSheetsService, EventHelper eventHelper)
+        public EventCommands(IEventsSheetsService eventsSheetsService)
         {
             this.eventsSheetsService = eventsSheetsService;
-            this.eventHelper = eventHelper;
         }
 
         [GroupCommand]
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -7,7 +7,6 @@ using DSharpPlus.CommandsNext;
 using DSharpPlus.CommandsNext.Attributes;
 using DSharpPlus.Entities;
 using DSharpPlus.Interactivity.Extensions;
-using static DiscordBot.Commands.EventHelper;
 
 namespace DiscordBot.Commands
 {
@@ -39,10 +38,12 @@ namespace DiscordBot.Commands
         };
 
         private readonly IEventsSheetsService eventsSheetsService;
+        private readonly EventHelper eventHelper;
 
-        public EventCommands(IEventsSheetsService eventsSheetsService)
+        public EventCommands(IEventsSheetsService eventsSheetsService, EventHelper eventHelper)
         {
             this.eventsSheetsService = eventsSheetsService;
+            this.eventHelper = eventHelper;
         }
 
         [GroupCommand]
@@ -277,7 +278,7 @@ namespace DiscordBot.Commands
 
             var signupsByResponse = await eventsSheetsService.GetSignupsByResponseAsync(eventKey);
 
-            var signupEmbed = GetSignupEmbed(discordEvent, signupsByResponse);
+            var signupEmbed = eventHelper.GetSignupEmbed(discordEvent, signupsByResponse);
 
             var signupMessage = await context.RespondAsync($"Signups are open for __**{discordEvent.Name}**__!", embed: signupEmbed);
             await eventsSheetsService.AddMessageIdToEventAsync(eventKey, signupMessage.Id);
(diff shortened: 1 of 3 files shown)


DiscordBot/Program.cs
```diff
@@ -40,9 +47,57 @@ private static async Task MainAsync()
             commands.RegisterCommands<EventCommands>();
 
             discord.UseInteractivity(new InteractivityConfiguration());
+            discord.MessageReactionAdded += OnMessageReactionAdded;
```
FridaTveit, line 50: Why `+=`? 🙂

DiscordBot/Program.cs
```diff
@@ -40,9 +47,57 @@ private static async Task MainAsync()
...
 
             await discord.ConnectAsync();
             await Task.Delay(-1);
         }
+
+        private static async Task OnMessageReactionAdded(
+            DiscordClient client,
+            MessageReactionAddEventArgs eventArguments)
+        {
+            if (eventArguments.User == client.CurrentUser)
```
FridaTveit, line 60: What's this case for? 🙂

DiscordBot/Program.cs
```diff
@@ -40,9 +47,57 @@ private static async Task MainAsync()
             commands.RegisterCommands<EventCommands>();
 
             discord.UseInteractivity(new InteractivityConfiguration());
+            discord.MessageReactionAdded += OnMessageReactionAdded;
 
             await discord.ConnectAsync();
             await Task.Delay(-1);
         }
+
+        private static async Task OnMessageReactionAdded(
```
FridaTveit, line 56: This method is a bit long, probably worth splitting it up? 🙂

DiscordBot/Program.cs
```diff
@@ -40,9 +47,57 @@ private static async Task MainAsync()
...
+            if (eventArguments.Emoji.GetDiscordName() == ":no_entry_sign:")
+            {
+                await eventsSheetsService.ClearResponsesForUserAsync(discordEvent.Key, eventArguments.User.Id);
+                await client.SendMessageAsync(dmChannel, $"You've signed off {discordEvent.Name}.");
+            }
+            else
+            {
+                try
+                {
+                    await eventsSheetsService.AddResponseForUserAsync(discordEvent.Key, eventArguments.User.Id,
```
FridaTveit, line 84: Each argument on a new line please

DiscordBot/Program.cs
```diff
@@ -40,9 +47,57 @@ private static async Task MainAsync()
...
+                        eventArguments.Emoji.GetDiscordName());
+                }
+                catch (ResponseNotFoundException)
+                {
+                    return;
+                }
+
+                await client.SendMessageAsync(dmChannel,
```
FridaTveit, line 92: Each argument on a new line please

DiscordBot/Program.cs
```diff
@@ -40,9 +47,57 @@ private static async Task MainAsync()
...
+                    $"You've responded to {discordEvent.Name} as {eventArguments.Emoji.Name}.");
+            }
+
+            var optionsList = await eventsSheetsService.GetSignupsByResponseAsync(discordEvent.Key);
```
FridaTveit, line 96: More descriptive name please

DiscordBot/Commands/EventHelper.cs
```diff
@@ -0,0 +1,95 @@
...
+using DiscordBot.DataAccess;
+using DiscordBot.DataAccess.Exceptions;
+using DiscordBot.DataAccess.Models;
+using DSharpPlus.Entities;
+
+namespace DiscordBot.Commands
+{
+    class EventHelper
+    {
+        private static IEventsSheetsService eventsSheetsService = new EventsSheetsService();
```
FridaTveit, line 13: I wouldn't expect this to have its own instance of the service. I'd expect it to either be passed in to each method or to the constructor for this class 🙂

DiscordBot/Commands/EventHelper.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using DiscordBot.DataAccess;
+using DiscordBot.DataAccess.Exceptions;
+using DiscordBot.DataAccess.Models;
+using DSharpPlus.Entities;
+
+namespace DiscordBot.Commands
+{
+    class EventHelper
```
FridaTveit, line 11: So how is this class different to EventCommands? Each class should have a clear responsibility and I'm not entirely sure what the responsibility of this class is? 🙂

FridaTveit (commented):
I've left some comments, otherwise looks good 🙂

DiscordBot/Program.cs
kwunkiuma: OnMessageReactionAdded is the handler for the MessageReactionAdded event. Is there a different way I should do this?

FridaTveit: I would expect just an assignment rather than `+=`? 🙂

DiscordBot/Program.cs
kwunkiuma: client.CurrentUser is the bot itself so this is the bot ignoring the reactions that are initialised on the signup sheet.

FridaTveit: Ah okay, might be worth a comment explaining that 🙂

DiscordBot/Program.cs
kwunkiuma: I get a compiler error with that:
"The event 'MessageReactionAdded' can only appear on the left hand side of += or -="

FridaTveit: I see! I guess it must be specific to the discord library api then. Is there anything in the docs about it? Either way, if the discord library enforces it then fair enough! 🙂

DiscordBot/Commands/EventHelper.cs
kwunkiuma: The methods in EventCommands aren't static and I can't reuse the methods within it without instantiating the class, which I don't think I'm supposed to do as the class is normally created from the API's RegisterCommands method.

kwunkiuma: Or I guess I could have made those methods static within EventCommands? I probably just didn't think of that.

FridaTveit: Ah okay, so these are just methods that need to be used both by the EventCommands and by the OnMessageReactionAdded method? That's probably fine. It would have been nice with a cleaner split of concerns but we don't want to make EventCommands too big either. So it's probably fine like this as long as you make sure they're using the same service instance 🙂

DiscordBot/Program.cs
kwunkiuma: Yup, there's an article on Events on the library website [here](https://dsharpplus.github.io/articles/beyond_basics/events.html).

DiscordBot/Program.cs
```diff
@@ -1,17 +1,25 @@
 ﻿using System;
+using System.Security.Cryptography.X509Certificates;
```
FridaTveit, line 2: Are you using this?

DiscordBot/Commands/EventHelper.cs
kwunkiuma: I was also thinking about extracting some methods from EventCommands into the helper method since EventCommands is quite big, what do you think?

DiscordBot/Program.cs
```diff
@@ -1,17 +1,25 @@
...
 using DSharpPlus.Interactivity.Extensions;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using static DiscordBot.Commands.EventHelper;
 
 namespace DiscordBot
 {
     internal class Program
     {
+        private static readonly IEventsSheetsService eventsSheetsService = new EventsSheetsService();
```
FridaTveit, line 21: You should only use one instance of the service

DiscordBot/Commands/EventHelper.cs
kwunkiuma: And since I'm relying on static methods I suppose I should pass the service as a parameter?

FridaTveit: Do the methods need to be static? Can the event handler method be an instance method?

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
```diff
@@ -40,10 +39,12 @@ internal class EventCommands : BaseCommandModule
         };
 
         private readonly IEventsSheetsService eventsSheetsService;
+        private readonly EventHelper eventHelper;
 
-        public EventCommands(IEventsSheetsService eventsSheetsService)
+        public EventCommands(IEventsSheetsService eventsSheetsService, EventHelper eventHelper)
         {
             this.eventsSheetsService = eventsSheetsService;
+            this.eventHelper = eventHelper;
```
FridaTveit: Is this used?

DiscordBot/Commands/EventHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+                return await eventsSheetsService.GetEventFromMessageIdAsync(messageId);
+            }
+            catch (EventNotFoundException)
+            {
+                return null;
+            }
+        }
+
+        public static Dictionary<ulong, IEnumerable<EventResponse>> GetResponsesByUser(
+            Dictionary<EventResponse, IEnumerable<ulong>> signupsByResponse)
```
FridaTveit, line 26: One argument per line

DiscordBot/Program.cs
```diff
@@ -40,9 +47,83 @@ private static async Task MainAsync()
...
             await discord.ConnectAsync();
             await Task.Delay(-1);
         }
+
+        private static async Task OnMessageReactionAdded(
+            DiscordClient client,
+            MessageReactionAddEventArgs eventArguments,
+            IEventsSheetsService eventsSheetsService)
+        {
+            if (eventArguments.User == client.CurrentUser)
```
FridaTveit, line 64: I still think a comment here would be useful

DiscordBot/Program.cs
```diff
@@ -40,9 +47,83 @@ private static async Task MainAsync()
...
+
+            var discordEvent = await GetEventFromMessageIdOrDefaultAsync(eventArguments.Message.Id, eventsSheetsService);
+            if (discordEvent == null)
+            {
+                return;
+            }
+
+            await ProcessReaction(client, eventArguments, discordEvent, eventsSheetsService);
+
+            await UpdateSignupSheet(eventArguments, discordEvent, eventsSheetsService);
```
FridaTveit, line 77: Can the two things above be done in parallel?

DiscordBot/Program.cs
```diff
@@ -40,9 +47,83 @@ private static async Task MainAsync()
...
+                        discordEvent.Key,
+                        eventArguments.User.Id,
+                        eventArguments.Emoji.GetDiscordName());
+                }
+                catch (ResponseNotFoundException)
+                {
+                    return;
+                }
+
+                await client.SendMessageAsync(
```
FridaTveit, line 109: This is incorrect in the response not found case

DiscordBot/Program.cs
```diff
@@ -40,9 +47,83 @@ private static async Task MainAsync()
...
+                    dmChannel,
+                    $"You've responded to {discordEvent.Name} as {eventArguments.Emoji.Name}.");
+            }
+        }
+
+        private static async Task UpdateSignupSheet(
```
FridaTveit, line 115: Is this actually updating the sign up sheet?

FridaTveit (commented):
I've left a few comments, let me know if any of them don't make sense 🙂

DiscordBot/Program.cs
kwunkiuma: I think the method should return if the response is not found?

DiscordBot/Program.cs
kwunkiuma: No because the updating the signup sheet obtains the new signup information from the sheet, meaning the sheet needs to be updated first.

DiscordBot/Program.cs
kwunkiuma: It refers to the sheet on the signup message, I'll rename it to UpdateSignupMessage to make it more clear.

DiscordBot/Commands/EventHelper.cs
kwunkiuma: There's only one argument, should I move the method name to the next line?

FridaTveit: Ah sorry, my mistake! Clearly struggling to read properly this afternoon 🤦

DiscordBot/Program.cs
FridaTveit: Ah okay 🙂

DiscordBot/Program.cs
FridaTveit: Yes, good point 🤦 Might be useful to print an error to the user in that case? 🙂

kwunkiuma: Any valid reaction will be automatically added by the bot so for a user to give an invalid reaction they will have had to manually choose an invalid one. The invalid reaction will still be removed by the bot so other users shouldn't mistake it for a valid reaction.

FridaTveit: Okay, sounds fine then 🙂

FridaTveit (approved):
Looks good 🙂

PR #42, opened 2020-11-18
Sheets api retries

benjimarshall:
Retry using an exponential backoff when we overload our Google Sheets 
API quota instead of crashing with an unhandled exception.

### DIFF
--- a/DiscordBot.DataAccess/SheetsServiceRequestsHelper.cs
+++ b/DiscordBot.DataAccess/SheetsServiceRequestsHelper.cs
@@ -147,8 +147,6 @@ namespace DiscordBot.DataAccess
                     {
                         retryMilliseconds *= 2;
                         retryMilliseconds += random.Next(1, 1000);
-
-                        Console.WriteLine($"Retry: {retryMilliseconds}");
                         await Task.Delay(retryMilliseconds);
                     }
                     else throw;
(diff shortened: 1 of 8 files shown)


DiscordBot.DataAccess/SheetsServiceRequestsHelper.cs
```diff
@@ -122,6 +127,36 @@ internal static ValueRange MakeCellUpdate(string range, object value)
...
+                    return await request.ExecuteAsync();
+                }
+                catch (GoogleApiException exception)
+                {
+                    if (exception.HttpStatusCode == HttpStatusCode.TooManyRequests)
+                    {
+                        retryMilliseconds *= 2;
+                        retryMilliseconds += random.Next(1, 1000);
+
+                        Console.WriteLine($"Retry: {retryMilliseconds}");
```
FridaTveit, line 151: Does this write anywhere sensible where you can see it if it's deployed? 🙂

FridaTveit (commented):
One question otherwise looks good 🙂

DiscordBot.DataAccess/SheetsServiceRequestsHelper.cs
benjimarshall: Whoops that was for testing! We haven't got any logging set up, so I probably should just remove this? 🙂

FridaTveit: Yes, that sounds sensible 🙂

benjimarshall: Removed in 8b6f7c6 🙂

FridaTveit (approved):
Looks good 🙂

PR #44, opened 2020-11-18
Wrap public methods in locks to ensure atomicity

benjimarshall:
The new `EventsSheetsService` seems a bit repetitive but I can't think of a better way of doing it. What do you think?

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -1,16 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using DiscordBot.DataAccess.Exceptions;
-using Google;
 using DiscordBot.DataAccess.Models;
-using Google.Apis.Auth.OAuth2;
-using Google.Apis.Services;
-using Google.Apis.Sheets.v4;
-using Google.Apis.Sheets.v4.Data;
-using static Google.Apis.Sheets.v4.SpreadsheetsResource.ValuesResource;
 
 namespace DiscordBot.DataAccess
 {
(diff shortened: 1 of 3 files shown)


DiscordBot.DataAccess/UnsafeEventsSheetsService.cs
```diff
@@ -0,0 +1,539 @@
...
+using DiscordBot.DataAccess.Models;
+using Google.Apis.Auth.OAuth2;
+using Google.Apis.Services;
+using Google.Apis.Sheets.v4;
+using Google.Apis.Sheets.v4.Data;
+using static Google.Apis.Sheets.v4.SpreadsheetsResource.ValuesResource;
+
+namespace DiscordBot.DataAccess
+{
+    public class UnsafeEventsSheetsService : IEventsSheetsService
```
FridaTveit, line 16: This can probably be internal if it's not meant to be used outside of this? 🙂

DiscordBot.DataAccess/UnsafeEventsSheetsService.cs
```diff
@@ -0,0 +1,539 @@
+﻿using System;
```
FridaTveit, line 1: This has just been moved from EventsSheetsService right, no other changes? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -40,527 +41,153 @@ public interface IEventsSheetsService
         Task<Dictionary<EventResponse, IEnumerable<ulong>>> GetSignupsByResponseAsync(int eventId);
     }
 
-    public class EventsSheetsService : IEventsSheetsService
+    public class EventsSheetsService : UnsafeEventsSheetsService
```
FridaTveit, line 44: I don't think an inheritance relationship is quite what you want here. Because you're not extending the functionality of the unsafe class, instead you're providing a wrapper for it. So I would recommend:
* Have both classes implement the sheets service interface. This will enforce the wrapper class implementing all the same methods as the inner class.
* Have the wrapper class contain an instance of the inner class which it will call methods on when safe
* Change the visibility of the inner class so that it can't be instantiated outside of this project

What do you think? 🙂

FridaTveit (commented):

DiscordBot.DataAccess/UnsafeEventsSheetsService.cs
benjimarshall: Yep! Changed in 81c8e0e 🙂

DiscordBot.DataAccess/UnsafeEventsSheetsService.cs
benjimarshall: Pretty much, The class name has changed, and some `virtual` keywords were added to make the inheritance work, but those have been removed again in 81c8e0e 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah I agree. I had considering both of these approaches and was a bit unsure of which to use. I probably should have checked that with you, but at least it didn't take me too long to switch it over. Changed to a wrapper class in 81c8e0e 🙂

FridaTveit (approved):
Looks good 🙂

PR #45, opened 2020-11-19
Change appearence of embeds

kwunkiuma:
Updated the preview for starting signups and result from the show command:
Before:
image

After:
image

The create event confirmation has also been changed but lacks the event key to conserve Sheets API calls.


FridaTveit (approved):
Looks great 🙂

PR #46, opened 2020-11-19
Add interface for batching event responses

benjimarshall:
Add a thread-safe buffer for storing reactions as they come in, and
process them up to every 10s.

Add an interface to the EventsSheetsService to process these batches
of requests. Add a model to store all information needed in a reaction:
a message (ie event) ID, a user ID and a reaction emoji (response
type).

I will implement the EventsSheetsService in another PR, and Frank
will plumb this into the Discord front end.

### DIFF
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -27,10 +27,8 @@ namespace DiscordBot.DataAccess
         Task<DiscordEvent> GetEventFromMessageIdAsync(ulong messageId);
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
 
-        Task AddResponseForUserAsync(int eventKey, ulong userId, string responseEmoji);
         Task AddResponseBatchAsync(IEnumerable<ResponseReaction> reactions);
         Task ClearResponseBatchAsync(IEnumerable<ResponseReaction> reactions);
-        Task ClearResponsesForUserAsync(int eventKey, ulong userId);
 
         Task<Dictionary<EventResponse, IEnumerable<ulong>>> GetSignupsByResponseAsync(int eventId);
     }
@@ -147,19 +145,6 @@ namespace DiscordBot.DataAccess
             }
         }
 
-        public async Task AddResponseForUserAsync(int eventKey, ulong userId, string responseEmoji)
-        {
-            try
-            {
-                await sheetsSemaphore.WaitAsync();
-                await eventsSheetsService.AddResponseForUserAsync(eventKey, userId, responseEmoji);
-            }
-            finally
-            {
-                sheetsSemaphore.Release();
-            }
-        }
-
         public async Task AddResponseBatchAsync(IEnumerable<ResponseReaction> reactions)
         {
             try
@@ -179,19 +164,6 @@ namespace DiscordBot.DataAccess
             {
                 await sheetsSemaphore.WaitAsync();
                 await eventsSheetsService.ClearResponseBatchAsync(reactions);
-            }
-            finally
-            {
-                sheetsSemaphore.Release();
-            }
-        }
-
-        public async Task ClearResponsesForUserAsync(int eventKey, ulong userId)
-        {
-            try
-            {
-                await sheetsSemaphore.WaitAsync();
-                await eventsSheetsService.ClearResponsesForUserAsync(eventKey, userId);
             }
             finally
             {
--- a/DiscordBot.DataAccess/EventsSheetsService.cs
+++ b/DiscordBot.DataAccess/EventsSheetsService.cs
@@ -28,6 +28,8 @@ namespace DiscordBot.DataAccess
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
 
         Task AddResponseForUserAsync(int eventKey, ulong userId, string responseEmoji);
+        Task AddResponseBatchAsync(IEnumerable<ResponseReaction> reactions);
+        Task ClearResponseBatchAsync(IEnumerable<ResponseReaction> reactions);
         Task ClearResponsesForUserAsync(int eventKey, ulong userId);
 
         Task<Dictionary<EventResponse, IEnumerable<ulong>>> GetSignupsByResponseAsync(int eventId);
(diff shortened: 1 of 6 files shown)


DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -28,6 +28,8 @@ public interface IEventsSheetsService
         Task<IEnumerable<DiscordEvent>> ListEventsAsync();
 
         Task AddResponseForUserAsync(int eventKey, ulong userId, string responseEmoji);
+        Task AddResponseBatchAsync(IEnumerable<ResponseReaction> reactions);
+        Task ClearResponseBatchAsync(IEnumerable<ResponseReaction> reactions);
         Task ClearResponsesForUserAsync(int eventKey, ulong userId);
```
FridaTveit, line 33: Do you still need the non-batch versions? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
```diff
@@ -158,6 +160,32 @@ public async Task AddResponseForUserAsync(int eventKey, ulong userId, string res
             }
         }
 
+        public async Task AddResponseBatchAsync(IEnumerable<ResponseReaction> reactions)
```
FridaTveit, line 163: Are these reactions meant to be all for the same message? Or could they be for different messages? 🙂

DiscordBot/Commands/ReactionBuffer.cs
```diff
@@ -0,0 +1,81 @@
...
+        private const int bufferDelayMilliseconds = 10 * 1000;
+
+        private readonly BlockingCollection<ResponseReaction> buffer;
+        private readonly SemaphoreSlim bufferStateLock;
+        private ReactionBufferState state;
+
+        private readonly IEventsSheetsService eventsSheetsService;
+        private readonly DiscordClient client;
+
+        public ReactionBuffer(IEventsSheetsService eventsSheetsService, DiscordClient client)
```
FridaTveit, line 29: I feel like this complication should be protected from the front end code if possible. I would expect the behaviour to be something like:
1. Users react to the message.
2. The reactions immediately appear on the message.
3. The reactions immediately get passed to the sheets service.
4. The sheets service only updates the google sheet in batches.

Since as far as I understand we're having to do this in batches due to the google sheets api restrictions. So it seems like it's a concern which should be hidden from the front end. What do you think? 🙂

FridaTveit (commented):
I've added a few comments, let me know if any of them don't make sense 🙂

DiscordBot/Commands/ReactionBuffer.cs
benjimarshall: The reason we did this was so that it could perform the Discord updates that need to happen after the reactions have been processed (deleting the reactions off the message, and updating the messages with the new signups). I do agree that it feels like we're mixing concerns, but I'm not sure if we can fully separate them due to the back-and-forth nature of operations between the two services. Do you have any ideas? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Yeah these reactions could be for any event message, from any user, and of any type. This service will sort them into the relevant groups and put it together in one big batch 🙂

DiscordBot/Commands/ReactionBuffer.cs
FridaTveit: Deleting/updating the reactions off the message seems like it could happen straight away without needing to wait for the sheet to be updated. What exactly do you mean by updating the messages with the new signups? 🙂

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: Nope, removed in 901aa51 🙂

benjimarshall: I'm going to add this back in, removing it breaks the discord bot. We will be able to replace it with the code @kwunkiuma is working on, but we can't remove it until then 🙂

DiscordBot/Commands/ReactionBuffer.cs
kwunkiuma: That's referring to the Discord message being updated to show the correct participants and their responses. This can only be done after the sheet has been updated we can only get updated information from the sheet.

benjimarshall: Yeah updating reactions happen first. 

The message for each event has a list of responses, and needs to be prodded to tell it to get the new version list and update the message. We don't want to do this as the reactions come in, because this would complicate how we process adding and clearing reactions together. Instead we can just decide that when the batch is processed, and the message just fetches the result. Sorry that wasn't clear in the first message 🙂

FridaTveit: Where exactly is this message and how is it being used? 🙂

kwunkiuma: The message that the bot sends for the signups.
It shows the list of participants and their responses.
image

DiscordBot.DataAccess/EventsSheetsService.cs
benjimarshall: I removed this again in 0ac5f7f because they are still being used. They won't be used after @kwunkiuma has finished his side of the interface, and then we will be able to remove them after that is pulled in 🙂

DiscordBot/Commands/ReactionBuffer.cs
FridaTveit: So this is the message which people react to? And it gets updated when people respond? Is that really necessary? 🙂

benjimarshall: Yes that is the message that people react to to respond to events. We think it's a really nice thing to have: by putting the event responses in this message normal users can see who has signed up already in a simple and neat way. This might be useful for more complex events, where the type of signup may depend on other signups (eg. positions for a football game).

Currently there isn't another good way of getting this information to general users, who don't have access to the underlying Google Sheet, and don't have permission to be posting commands in the bot-command channel. We would rather not add them to this channel, because that would begin to break down the divide between the bot-admin roles and the normal users. What do you think? 🙂

FridaTveit: Can't you see who added the reaction by hovering over it like you can in slack?

benjimarshall: At the moment the reactions are being deleted as part of processing the reaction. This is a fairly common practice with Discord bots; it allows the bot to track which work has and hasn't been processed (eg. allowing it to catch up on work that was missed when the bot was turned off). We haven't implemented this yet but with this design it would not result in any behaviour change if we were to implement this.

The process at the moment roughly is: 1) react to a message 2) process the consequences of this reaction 3) remove the reaction, notify the user they have signed up, and update the message.

Part of the reason we chose to do this was that subscribing to reaction removals was much harder in the previous version of DSharpPlus, but could be easier now. We also vaguely remember talking about this in the meeting last Friday with Gareth, but we can't quite remember what was said. We could change it over to having persistent reactions for responses if you prefer 🙂

FridaTveit: I think that if updating the message is making the implementation significantly more complicated (which it looks like it is), then it's worth reconsidering whether it's something we need to do. It's not something I would necessarily expect a bot like this to do. You could check with Gareth what he thinks 🙂

benjimarshall:
@FridaTveit I think this is ready for another review 🙂 @kwunkiuma has been working on plumbing this into the Discord API (to update the event messages on processing the reactions, and to send the reactions to this buffer), and I am finishing off implementing the batch update methods in the `EventsSheetsService` that were defined here. 🙂

FridaTveit (approved):
Looks good 🙂

PR #48, opened 2020-11-19
Add response set options on event creation

SChandarana:
You can now choose between a set of hard-coded response sets for your signup as demonstrated in the image below

image

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -161,16 +161,14 @@ namespace DiscordBot.Commands
                 Title = "Response options"
             };
 
-
-            var i = 0;
-            foreach (var responseSet in ResponseSets)
+            for (var i = 0; i < ResponseSets.Length; i++)
             {
+                var responseSet = ResponseSets[i];
                 var responseSetString = string.Join(
                     ", ",
                     responseSet.Select(response => $"{response.ResponseName} - {response.Emoji}")
                 );
                 responseEmbed.AddField($"Response set {i}", responseSetString);
-                i += 1;
             }
 
             await context.RespondAsync(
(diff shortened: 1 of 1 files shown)


SChandarana:
Ah i was wondering why it still showed the previous commits, it's due to the squash. Is it still okay to leave this PR up or do you want me to fix it to branch off of the current main (to remove the extra commits)

DiscordBot/Commands/EventCommands.cs
```diff
@@ -130,13 +156,55 @@ public async Task CreateEvent(CommandContext context, string eventName, string e
...
         {
-            await eventsSheetsService.AddEventAsync(eventName, eventDescription, eventTime);
+            var responseEmbed = new DiscordEmbedBuilder
+            {
+                Title = "Response options"
+            };
+
+
+            var i = 0;
+            foreach (var responseSet in ResponseSets)
```
FridaTveit, line 166: If you use a for loop instead of foreach you don't need to keep track of a separate index 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -130,13 +156,55 @@ public async Task CreateEvent(CommandContext context, string eventName, string e
...
 
+        [Command("create")]
+        public async Task CreateEvent(
+            CommandContext context,
+            string eventName,
+            string eventDescription,
+            DateTime eventTime,
+            int? responseSetIndex)
+        {
+            if (responseSetIndex == null || responseSetIndex.Value >= ResponseSets.Length)
```
FridaTveit, line 192: What if it's a negative number?

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
SChandarana: Good point!

DiscordBot/Commands/EventCommands.cs
```diff
@@ -161,16 +161,14 @@ public async Task CreateEvent(CommandContext context, string eventName, string e
                 Title = "Response options"
             };
 
-
-            var i = 0;
-            foreach (var responseSet in ResponseSets)
+            for (var i = 0; i < ResponseSets.Length; i++)
             {
+                var responseSet = ResponseSets[i];
```
FridaTveit, line 166: Very minor but you could just inline this? You're only using it once as far as I can tell so no real advantage to having a variable 🙂

FridaTveit (commented):
One tiny thing, otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #50, opened 2020-11-20
Change list to only show upcoming events

kwunkiuma:
Listing all events will only show events that are occurring in the future.

Events are also ordered chronologically

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -279,20 +279,24 @@ namespace DiscordBot.Commands
         {
             var eventsList = await eventsSheetsService.ListEventsAsync();
 
+            var sortedEventsList = eventsList
+                .Where(x => x.Time > DateTime.Now)
+                .OrderBy(x => x.Time);
+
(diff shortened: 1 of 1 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -279,20 +279,24 @@ public async Task ListEvents(CommandContext context)
         {
             var eventsList = await eventsSheetsService.ListEventsAsync();
 
+            var sortedEventsList = eventsList
+                .Where(x => x.Time > DateTime.Now)
```
kwunkiuma, line 283: Might it be better to show events that happened earlier on the same day?

kwunkiuma (commented):
Also have one question.

DiscordBot/Commands/EventCommands.cs
FridaTveit: No, I think all upcoming events is the most useful. It's always possible to add a separate command for listing all events at some point if necessary 🙂

FridaTveit (commented):
Looks good 🙂

PR #52, opened 2020-11-20
Send signup message to specific channel

kwunkiuma:
Starting a signup will now send the signup message to the channel named events.

If the channel doesn't exist then the users is informed of the error.

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -17,6 +17,9 @@ namespace DiscordBot.Commands
     internal class EventCommands : BaseCommandModule
     {
         public const string ClearReaction = ":no_entry_sign:";
+        public const string RefreshReaction = ":arrows_counterclockwise:";
+
+        public const string SignupChannelName = "events";
 
         private static readonly string[] EventOperations =
         {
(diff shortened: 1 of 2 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -381,18 +384,32 @@ private async Task CreateSignupSheet(CommandContext context, int eventKey)
                 return;
             }
 
-            await SendSignupMessage(context, eventKey);
+            try
+            {
+                var signupChannel = context.Guild.Channels
+                    .First(channel => channel.Value.Name == SignupChannelName)
```
FridaTveit, line 390: Minor but you could probably use `FirstOrDefault` here instead and avoid the try/catch (instead check for null). Makes the code less nested and a bit easier to read I think 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -17,6 +17,9 @@ namespace DiscordBot.Commands
     internal class EventCommands : BaseCommandModule
     {
         public const string ClearReaction = ":no_entry_sign:";
+        public const string RefreshReaction = ":arrows_counterclockwise:";
+
+        public const string SignupChannelName = "events";
```
FridaTveit, line 22: Do these new consts need to be public? 🙂

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
kwunkiuma: RefreshReaction should have been used in Program.cs but SignupChannelName can be private.

DiscordBot/Commands/EventCommands.cs
kwunkiuma: Done!

DiscordBot/Commands/EventCommands.cs
```diff
@@ -384,18 +384,17 @@ private async Task CreateSignupSheet(CommandContext context, int eventKey)
...
             }
 
-            try
-            {
-                var signupChannel = context.Guild.Channels
-                    .First(channel => channel.Value.Name == SignupChannelName)
-                    .Value;
+            var signupChannel = context.Guild.Channels
+                .FirstOrDefault(channel => channel.Value.Name == SignupChannelName)
+                .Value;
```
FridaTveit, line 389: I don't think you can call `Value` here if null has been returned? Have you tested this for a channel which doesn't exist? 🙂

FridaTveit (commented):
One question otherwise looks good 🙂

DiscordBot/Commands/EventCommands.cs
kwunkiuma: I had that same thought and wanted to use ?.Value but the IDE didn't allow it and it appears to be working when the channel doesn't exist.
image

kwunkiuma: It seems the default value is a key value pair of [0, null].

FridaTveit: Ah I see 🙂

FridaTveit (approved):
Looks good 🙂

PR #53, opened 2020-11-20
Update readme with todo list

SChandarana:
Added some useful links (benji will be adding some more things re: google sheets) and also a nice Checklist of things we have done/ still need to do. Let me know if theres anything else you think of adding :D


FridaTveit (approved):
Looks good 🙂

PR #54, opened 2020-11-20
Add Google Sheets instructions to the README

benjimarshall:


README.md
```diff
@@ -76,9 +80,23 @@ have the ability to add the bot to a server in which you have the "Manage Server
...
+1. Add yourself as an editor to the softwire-discord-bot project on Google Cloud Platform, or get someone else who is an
+editor to follow the subsequent steps.
+    1. Login into the "Google Developers Project Manager" account (in the Zoho chamber)
+    2. In IAM & Admin -> IAM, add your Google account as an editor
+2. In IAM & Admin -> Service accounts, create a new service account
+    1. Name the account with some reference to your name ideally
+    2. Don't give it any access, roles, or give users special access (so just click continue and done)
+3. Click on your newly made service account, and add a key (Add key -> create new key -> JSON -> Create)
+4. A JSON credentials file will download. You will need this to run the bot locally.
+5. Create a Google Sheet of the right format (easiest just to copy someone else's test sheet)
```
FridaTveit, line 92: Is it possible to give access to a test sheet to copy? So they don't need to track down a previous contributor to find one 🙂

FridaTveit (commented):
One question, otherwise looks great 🙂

README.md
benjimarshall: Yeah that would be possible. I think we discussed this in one of the previous PRs, the issue with this is that if multiple people tried to use the test sheet at once it has the potential to misbehave and fail mysteriously because the `EventsSheetsService` expects that it is the only thing editing the sheet. I could add a test sheet with some caveats and warnings if you like? 🙂

benjimarshall: Wait I misread this, sorry! I'll add that in now 🙂

benjimarshall: Added in f3738e2 🙂

FridaTveit (approved):
Looks great 🙂

PR #55, opened 2020-11-20
Update front end to batch responses

kwunkiuma:
The front end now uses the batched reaction handling.

The message once again refreshes automatically on signups.

Managed to avoid adding the client to the buffer with the trade-off of removing DM confirmations, which would have been difficult to implement in batches anyways as invalid responses are silently ignored.

### DIFF
--- a/DiscordBot/Program.cs
+++ b/DiscordBot/Program.cs
@@ -3,10 +3,8 @@ using System.Threading.Tasks;
 using DiscordBot.Commands;
 using DiscordBot.DataAccess;
 using DiscordBot.DataAccess.Exceptions;
-using DiscordBot.DataAccess.Models;
 using DSharpPlus;
 using DSharpPlus.CommandsNext;
-using DSharpPlus.Entities;
 using DSharpPlus.EventArgs;
 using DSharpPlus.Interactivity;
 using DSharpPlus.Interactivity.Extensions;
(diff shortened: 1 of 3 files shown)


DiscordBot/Program.cs
```diff
@@ -47,102 +45,75 @@ private static async Task MainAsync()
...
-                await OnMessageReactionAdded(client, eventArguments, eventsSheetsService);
+                await OnMessageReactionAdded(client, eventArguments, eventsSheetsService, reactionBuffer);
             };
 
             await discord.ConnectAsync();
             await Task.Delay(-1);
         }
 
-        private static async Task OnMessageReactionAdded(
+        private static Task OnMessageReactionAdded(
```
FridaTveit, line 60: Does this need to return a Task at all, or could it just have void return type? 🙂

DiscordBot/Program.cs
```diff
@@ -47,102 +45,75 @@ private static async Task MainAsync()
...
+            // Skip if reaction is not in the signup channel
+            if (eventArguments.Channel.Name != EventCommands.SignupChannelName)
             {
-                return;
+                return Task.CompletedTask;
             }
 
-            await ProcessReaction(client, eventArguments, discordEvent, eventsSheetsService);
+            _ = Task.Run(async () =>
+            {
```
FridaTveit, line 79: Does this need to be inside a block? 🙂

DiscordBot/Program.cs
```diff
@@ -47,102 +45,75 @@ private static async Task MainAsync()
...
             MessageReactionAddEventArgs eventArguments,
-            DiscordEvent discordEvent,
             IEventsSheetsService eventsSheetsService)
         {
-            var signupsByResponse = await eventsSheetsService.GetSignupsByResponseAsync(discordEvent.Key);
-            await eventArguments.Message.ModifyAsync(
-                eventArguments.Message.Content,
-                GetSignupEmbed(discordEvent, signupsByResponse).Build()
-            );
+
```
FridaTveit, line 107: Unneeded empty line?

FridaTveit (commented):

DiscordBot/Program.cs
kwunkiuma: Nope, removed.

FridaTveit (approved):
Looks good 🙂

DiscordBot/Program.cs
benjimarshall: I've just had a look at this with @kwunkiuma. Basically, the `discord.MessageReactionAdded` event handler on line 51 expects to receive a `Task`, but there's no need for `OnMessageReactionAdded` to returning one in all cases. I've changed it so it returns `void` and the hander here returns a `Task.CompletedTask` in 8c81656 🙂

PR #56, opened 2020-11-20
Add Descriptions to commands and field for use with ?help

SChandarana:
I've added descriptions to all the commands and fields. Unfortunately I couldn't find a way to hide the multiple versions of CreateEvent etc so i had to just duplicate the descriptions. This is to avoid a flood of "No Description Provided" when using `?help event create`

### DIFF
--- a/DiscordBot/Commands/EventCommands.cs
+++ b/DiscordBot/Commands/EventCommands.cs
@@ -14,6 +14,7 @@ namespace DiscordBot.Commands
 {
     [Group("event")]
     [RequireRoles(RoleCheckMode.All, "Bot Whisperer")]
+    [Description("Initiates the wizard for event-related actions, use sub commands to bypass these")]
     internal class EventCommands : BaseCommandModule
     {
         public const string ClearReaction = ":no_entry_sign:";
@@ -76,7 +77,6 @@ namespace DiscordBot.Commands
         }
 
         [GroupCommand]
-        [Description("Initiates the wizard for event-related actions")]
         public async Task Event(CommandContext context)
         {
             await context.RespondAsync(
@@ -119,6 +119,7 @@ namespace DiscordBot.Commands
         }
 
         [Command("create")]
+        [Description("Begins the creation wizard")]
         public async Task CreateEvent(CommandContext context)
         {
             await context.RespondAsync($"{context.Member.Mention} - what is the name of the event?");
@@ -132,7 +133,9 @@ namespace DiscordBot.Commands
         }
 
         [Command("create")]
-        public async Task CreateEvent(CommandContext context, string eventName)
+        public async Task CreateEvent(
+            CommandContext context,
+            [Description("Must be surrounded in quotes")] string eventName)
         {
             await context.RespondAsync($"{context.Member.Mention} - give an event description.");
             var eventDescription = await GetUserResponse(context);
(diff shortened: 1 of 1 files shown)


DiscordBot/Commands/EventCommands.cs
```diff
@@ -132,7 +133,9 @@ public async Task CreateEvent(CommandContext context)
         }
 
         [Command("create")]
-        public async Task CreateEvent(CommandContext context, string eventName)
+        public async Task CreateEvent(
+            CommandContext context,
+            [Description("Must be surrounded in quotes")] string eventName)
```
FridaTveit, line 138: Surrounded by?

DiscordBot/Commands/EventCommands.cs
```diff
@@ -225,7 +237,7 @@ public async Task RemoveEvent(CommandContext context)
         }
 
         [Command("remove")]
-        public async Task RemoveEvent(CommandContext context, int eventKey)
+        public async Task RemoveEvent(CommandContext context, [Description("Optional key corresponding to event")] int eventKey)
```
FridaTveit, line 240: It's not really optional, you will need to provide the key at some point right? 🙂

DiscordBot/Commands/EventCommands.cs
```diff
@@ -340,10 +355,10 @@ private async Task EditEvent(CommandContext context, int eventKey)
...
         [Command("edit")]
         private async Task EditEvent(
-            CommandContext context, 
-            int eventKey, 
-            string editField,
-            DiscordEmbedBuilder? eventEmbed = null)
+            CommandContext context,
+            [Description("Optional key corresponding to event")] int eventKey,
+            [Description("'name', 'description' or 'time'")] string editField,
+            [Description("Ignore this")] DiscordEmbedBuilder? eventEmbed = null)
```
FridaTveit, line 361: Does this actually get printed?

FridaTveit (commented):

DiscordBot/Commands/EventCommands.cs
SChandarana: The description shows how the command works, you can still invoke the command without the event key (and then it will subsequently prompt you for it)

DiscordBot/Commands/EventCommands.cs
SChandarana: Yep unfortunately ?help event edit will pop up all the fields including the embedbuilder field. I just decided for now it would be best to add _some_ kind of description rather than nothing

DiscordBot/Commands/EventCommands.cs
FridaTveit: I still think saying it's optional is misleading though so I would leave that out 🙂

SChandarana: I can easily just remove the "optional" if you think it would make more sense that way though :D

DiscordBot/Commands/EventCommands.cs
FridaTveit: Hm, that's not great... Ideally we'd find a way to make it not print it as an option at all. But as we don't have much time left I'd suggest changing the description to something like "Internal bot use only, please ignore" 🙂

SChandarana: Yep that's fine, the reason this is the case is that we're using the default help command. Ideally we'd want to use our own help command which is formatted in a way we want it to be

FridaTveit (approved):
Looks good 🙂

